PR #4, opened 2024-12-31
Fix ReadTheDocs Prebuild

JacobKnightley:


JacobKnightley:
Also renaming functions that were improperly showing as public functions.

PR #8, opened 2025-01-02
Authentication, Logging, & Documentation

JacobKnightley:
This pull request focuses on improving logging, adding sample scripts, and updating the authentication approach to support users passing in their own authentication. 

- Documentation Updates
- Sample Scripts added to document how to use the workspace object, how to pass in custom authentication, and how to define variables based on target environments
- Authentication updated to support credential passed into workspace object, or default to current behavior if not provided
- Logging added throughout the module to write to a common log file.  Most print statements were changed to log statements.  Additionally added logging outputs for the full API calls for better debugging of unhandled exceptions.


JacobKnightley:
This pull request focuses on improving logging, adding sample scripts, and updating the authentication approach to support users passing in their own authentication. 

- Documentation Updates
- Sample Scripts added to document how to use the workspace object, how to pass in custom authentication, and how to define variables based on target environments
- Authentication updated to support credential passed into workspace object, or default to current behavior if not provided
- Logging added throughout the module to write to a common log file.  Most print statements were changed to log statements.  Additionally added logging outputs for the full API calls for better debugging of unhandled exceptions.

JacobKnightley:
Also consolidated dependency on flake8/black to use ruff instead.

PR #42, opened 2025-01-18
Introduce retry with backoff

jmuziki:
This pull request includes several changes to the `invoke` method in the `src/fabric_cicd/_common/_fabric_endpoint.py` file, focusing on improving error handling and retry logic for API requests. The most important changes include the addition of a new `_handle_retry` method, enhanced error messaging, and improved handling of API throttling and long-running operations.

Improvements to error handling and retry logic:

* Added a new `_handle_retry` method to handle retry logic with exponential backoff based on the response headers. This method is called when the `Retry-After` header is present in the response.
* Enhanced error messaging for failed operations and deployment attempts, improving readability and clarity. [[1]](diffhunk://#diff-df48b77e163b4a13c97f80408ad6640fca48ffdc60c19e085b4c73589933178cL76-R88) [[2]](diffhunk://#diff-df48b77e163b4a13c97f80408ad6640fca48ffdc60c19e085b4c73589933178cL129-R160)

Handling of API throttling and long-running operations:

* Improved handling of API throttling by using the `_handle_retry` method when the `Retry-After` header is present in the response. This change ensures that the retry logic is consistent and robust.
* Added logic to handle retries for long-running operations when the `Retry-After` header is present in the response. This change ensures that the system waits appropriately before retrying the operation.

Minor improvements:

* Fixed a typo in a log message to improve clarity.

### DIFF
--- a/src/fabric_cicd/_common/_fabric_endpoint.py
+++ b/src/fabric_cicd/_common/_fabric_endpoint.py
@@ -49,7 +49,6 @@ class FabricEndpoint:
                         "Content-Type": "application/json; charset=utf-8",
                     }
                     response = requests.request(method=method, url=url, headers=headers, json=body)
-
                 else:
                     headers = {"Authorization": f"Bearer {self.aad_token}"}
                     response = requests.request(method=method, url=url, headers=headers, files=files)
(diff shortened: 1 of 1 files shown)


src/fabric_cicd/_common/_fabric_endpoint.py
```diff
@@ -37,11 +37,11 @@ def invoke(self, method, url, body="{}", files=None):
...
         :return: A dictionary containing the response headers, body, and status code.
         """
-        exit_loop = False
-        iteration_count = 0
+        max_retries = 5
+        base_delay = 0.5
         long_running = False
 
-        while not exit_loop:
+        for attempt in range(max_retries):
```
JacobKnightley: One thing to consider here is that we only want to use max tries if we're pulsing a long running operation or we have a reserved name, or are being throttled.  And, for something like deploying a semantic model which will go into the long operation it could technically be 5-10 minutes so defining a max retry could be risky.  

We may realistically need to break this up into multiple functions.  an api call where we actually run the call and parse errors.  Then if response is of "type" retry, then we go into a separate loop.  I.e. invoke first, if long running, step into loop, invoke again until end state.

shirasassoon (approved):

Issue #62, opened 2025-01-28
Bug Report


chaosmonkey86:
### Library Version

0.1.2

### What is the problem?

`base_delay` in `_check_environment_publish_state` needs to be a float, otherwise when it retries in the event of a "pending" response, you get `AttributeError: 'int' object has no attribute 'is_integer'`

![Image](https://github.com/user-attachments/assets/f48ff1de-0198-40ca-bea7-b61415649e10)

### Is there an ideal solution?

Change to 5.0

JacobKnightley:
Thanks for reporting!  Can you share in which scenario this occurred?  Mainly from an integration testing perspective want to make sure we're stepping in and hitting this

JacobKnightley:
Same as #63

PR #66, opened 2025-01-28
Fixed integer check

### LINKED ISSUE #63: handle_retry function failure due to AttributeError
### Library Version

0.1.2

### What is the problem?

When publishing Fabric `Environment` artifact (`"type": "Environment"`), the `_check_environment_publish_state` function will be called which performs `handle_retry` function:

Image

with `base_delay=5`, this fails with the following error message:

Image

This happens because `base_delay=5` is of `int` type and therefore does not have `is_integer()` method.

1. Python version >= 3.12
`int.is_integer()` was introduced in >= Python version 3.12, so changing the Python version to >= 3.12 solves the issue. See [https://docs.python.org/3/library/stdtypes.html#int.is_integer](https://docs.python.org/3/library/stdtypes.html#int.is_integer)
2. Python version < 3.12
If Python version is kept < 3.12, the following change can be done on `handle_retry` function.
Add `base_delay = float(base_delay)` `int` conversion to `float`:
``` python
retry_after = float(response_retry_after)
base_delay = float(base_delay)
delay = min(retry_after, base_delay * (2**attempt))
```

_No response_

vytautas9:
Solves an issue with `int.is_integer()` not being available in Python versions < 3.12.

closes #63

### DIFF
--- a/src/fabric_cicd/_common/_fabric_endpoint.py
+++ b/src/fabric_cicd/_common/_fabric_endpoint.py
@@ -242,6 +242,7 @@ def handle_retry(attempt, base_delay, max_retries, response_retry_after=60, prep
     """
     if attempt < max_retries:
         retry_after = float(response_retry_after)
+        base_delay = float(base_delay)
         delay = min(retry_after, base_delay * (2**attempt))
 
         # modify output for proper plurality and formatting


JacobKnightley (approved):

JacobKnightley:
Fixes #63, Fixes #62

PR #79, opened 2025-01-31
Fixes #76 - Continue on Unpublish failure

### LINKED ISSUE #76: Unpublish Step trying to Delete Default Semantic Model
### Library Version

0.1.2

### What is the problem?

When running unpublish with item type Semantic Model specified, it tries to delete the default Semantic Model.  Default Semantic Models should not ever be deleted, only the lakehouse can be deleted. 

####################################################################################################
####################################################################################################

[INFO] 16:35:22 - Unpublishing SemanticModel 'test'
[ERROR] 16:35:22 - Unhandled error occurred calling DELETE on 'https://msitapi.fabric.microsoft.com//v1/workspaces//items/'. Message: Target entity does not support the required operation

Response Status: 400
Response Headers:
{
    "Cache-Control": "no-store, must-revalidate, no-cache",
    "Pragma": "no-cache",
    "Transfer-Encoding": "chunked",
    "Content-Type": "application/json; charset=utf-8",
    "x-ms-public-api-error-code": "BadRequest",
    "X-Frame-Options": "deny",
    "X-Content-Type-Options": "nosniff",
    "Access-Control-Expose-Headers": "RequestId",
    "Date": "Fri, 31 Jan 2025 00:35:21 GMT"
}
Response Body:
{"requestId":"","errorCode":"BadRequest","message":"Target entity does not support the required operation"}

For now, we should simply continue and raise a warning.  Need to work with product to determine if there's a way to identify default semantic model item type and not include in the delete process.  

_No response_

JacobKnightley:
This pull request includes a change to the error handling mechanism in the `_unpublish_item` method of the `fabric_workspace.py` file. The change adds a try-except block to catch exceptions during the deletion of an item and logs a warning message if an exception occurs.

Error handling improvement:

* [`src/fabric_cicd/fabric_workspace.py`](diffhunk://#diff-3bd0f70ed06c7fc7a0f77378aecc6fb108eeb5263161db3b22270f3de4df99c0R454-R458): Added a try-except block to handle potential exceptions during the deletion of an item and log a warning message with the item type and name if an exception occurs.


JacobKnightley:
Note that this is a temporary fix.  Need to find permanent fix to identify what items are "default" items and can't be deleted.

jmuziki (approved):

PR #162, opened 2025-03-18
Resolves #166 Support Environment Variables in Parameter.yml

richbenmintz:
Modified the _refresh_parameter_file methond in fabric_workspace.py

- if feature flag 'enable_devops_variables' is set
  - load parameter file as dict
    - for each item in the find_replace list
      - update the value for self.environment with the os.environment[self.environment]
 - assign the modified dict to self.environment_parameter

I modified the hello world notebook to include a lakehouse dependency and included the strings in the os.environs in debug.local.py

`append_feature_flag("enable_devops_variables")`
`os.environ["sql_con_ppe_var"] = "104e7ff8-70c7-4e84-a6ad-e4e1f4ecd1b5"`
`os.environ["lakehouse_ppe_var"] = "104e7ff8-70c7-4e84-a6ad-e4e1f4ecd1b5"`
`os.environ["lakehouse_ppe_workspace"] = "f0e3fa10-e2d7-4ce6-a508-4e6e16ba3a27"`

`"lakehouse": {`
`# META       "default_lakehouse": "replace_lakehouse_id",`
`# META       "default_lakehouse_name": "lh_test",`
`# META       "default_lakehouse_workspace_id": "replace_lakehouse_workspace"`
`# META     }`

image

### DIFF
--- a/docs/how_to/parameterization.md
+++ b/docs/how_to/parameterization.md
@@ -2,6 +2,8 @@
 
 To handle environment-specific values committed to git, use a `parameter.yml` file. This file supports programmatically changing values based on the `environment` field passed into the `FabricWorkspace` object. If the environment value is not found in the `parameter.yml` file, any dependent replacements will be skipped. This file should sit in the root of the `repository_directory` folder specified in the FabricWorkspace object.
 
+**Important Notice:** The `parameter.yml` file structure has been recently updated. Please refer to the documentation below for important changes. There is a grace period from **March 24, 2025** to **April 24, 2025** during which the old structure will still be supported, allowing users to migrate to the new structure.
+
 Example of parameter.yml location based on provided repository directory:
 
 ```python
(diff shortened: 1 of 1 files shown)


src/fabric_cicd/fabric_workspace.py
```diff
@@ -129,13 +129,22 @@ def __init__(
...
+
         parameter_file_path = self.repository_directory / "parameter.yml"
         self.environment_parameter = {}
-
+        #
         if parameter_file_path.is_file():
             logger.info(f"Found parameter file '{parameter_file_path}'")
             with parameter_file_path.open(encoding="utf-8") as yaml_file:
-                self.environment_parameter = yaml.safe_load(yaml_file)
+                if "enable_devops_variables" in feature_flag:
```
JacobKnightley: We should probably use a more generic name since this isn't ADO specific, just environment variables.  

Additionally, we should consider whether we should explicitly use a syntax so that only keys with something like $ENV: are replaced.  That way if we have a value that just happens to align with an environment variable name we don't accidentally replace

JacobKnightley (changes requested):
Remove changes to files outside of the src files.  Also we will need this documented in the docs section of the repo, under how to for the feature flags

richbenmintz:
made the requested code changes

JacobKnightley:
@richbenmintz  I agree with the changes, we will hold off until @shirasassoon  has her changes landed and then you can rebase and land in the new module she's created.

richbenmintz:
Thanks Jacob, appreciate your support

shirasassoon:
@richbenmintz The changes have been landed.

richbenmintz:
Should I be updating the _validate_load_parameters_to_dict method, and adding the dict replace functionality there?

shirasassoon:
@richbenmintz You can do that, or you can add your logic as a function to _utils.py in the _parameter module, import it into the class file and then call it within validate_load_parameters_to_dict method to keep it more contained. 

Please add documentation for this change. Thanks!

PR #224, opened 2025-04-08
updates to support, variable library support in parameters and a some…

richbenmintz:
I have added the ability to define cicd parameterization for Variable Libraries, and a somewhat generic approach to key value replacement for item definitions that are defined as json. 

the yaml for variable libs would look like

`variable_libraries:
  - PPE:
      - library_name: dsi-vars
        variables:
          - name: var_test
            note: ''
            type: String
            value: test_value_again
          - name: var_string
            note: ''
            type: String
            value: test_string_var`

PR is lacking documentation updates, but wanted to get it in front of you to see if it was something you wanted to pursue, otherwise I will just use it myself

### DIFF
--- a/src/fabric_cicd/_parameter/_parameter.py
+++ b/src/fabric_cicd/_parameter/_parameter.py
@@ -36,6 +36,10 @@ class Parameter:
             "maximum": {"instance_pool_id", "replace_value", "item_name"},
         },
         "spark_pool_replace_value": {"type", "name"},
+        "variable_libraries": {
+            "minimum": {"library_name", "variables"},
+            "maximum": {"library_name", "variables"},
+        },
     }
 
     def __init__(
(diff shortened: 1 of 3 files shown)


richbenmintz:
updated yaml with alternate set support
`variable_libraries:       
 - PPE:
    - library_name: "dsi-vars"
      variables:
        - name: var_test
          note: 'note_test'
          type: String
          value: test_value_again_again
        - name: var_string
          note: 'testing_notes'
          type: String
          value: test_string_var_testing
      alternate_sets: 
       - set_name: "alt_set"
         variables:
           - name: "var_test"
             value: "alternate_value"`

richbenmintz:
Couple use Case for why parameterizing variable libraries would make sense as opposed to managing many alternate sets:
- To many alternate sets to manage effectively in the UI
- deploying N number of workspaces for data consumers, where you want to provide them with a set of variables for them to do their work effectively, however the variable values are not the same for each workspace, if you have a base workspace item library, you could then deploy to workspaces with variable group values that are contextual to the workspace being deployed to.
  - I suppose the counter argument would be that is more of an infra CLI type example and the scripts would create the variable library with the appropriate values.
Just some thoughts

JacobKnightley:
Hi @richbenmintz  Shira is now back and has merged her most recent changeset.  If you can post a new github issue with exactly what your ideal scenario is, we can land on what makes sense to include in fabric-cicd.  And then cherry pick the features in this branch to what is agreed on

richbenmintz:
Sure thing thanks

JacobKnightley:
Closing due to aging.

Issue #200, opened 2025-04-01
List workspaces / get workspace id from name


PLsergent:
### What is the feature?

We currently need to specify a `workspace_id` to convey a deployment. Would it be interesting to add a method in the `FabricEndpoint` class (or somewhere else) to list the workspaces, and/or, get a workspace id from a workspace name? Maybe even directly a `workspace_name` param in the `FabricWorkspace` class. What are your thoughts?

### Additional context

I quickly wrote a function that I'll use in my code based on what I saw in the `FabricEndpoint` class, but I guess we can make it a method directly in the class:

```python
def get_workspace_id_from_name(workspace_name, token_credential, base_api_url="https://api.fabric.microsoft.com/v1"):
    endpoint = FabricEndpoint(
        # if credential is not defined, use DefaultAzureCredential
        token_credential=(
            DefaultAzureCredential() if token_credential is None else validate_token_credential(token_credential)
        )
    )
    # > make it a method in FabricEndpoint?
    response = endpoint.invoke(method="GET", url=f"{base_api_url}/workspaces")
    for item in response["body"]["value"]:
        if item["displayName"] == workspace_name:
            return validate_workspace_id(item["id"])
    return None
```

JacobKnightley:
@PLsergent  Thanks for the feature suggestion!  This should be fairly straight forward to implement.  We'll add this to our backlog

PR #239, opened 2025-04-13
Added sample to dynamically lookup workspace id using name

hurtn:
Example demonstrating:  
1. Access variable group values from Python. Note for sensitive variables ensure the variable group is linked to key vault. See https://learn.microsoft.com/en-us/azure/devops/pipelines/library/link-variable-groups-to-key-vaults?view=azure-devops
2. Use of Service Principal Name (SPN) with a Secret credential flow, leveraging the ClientSecretCredential class. 
3. Use the Fabric reset APIs to lookup the workspace ID based on workspace name
4. Using debug log level

### DIFF
--- a/sample/dynamically_lookup_workspace_id_using_name.py
+++ b/sample/dynamically_lookup_workspace_id_using_name.py
@@ -1,0 +1,103 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""
+Example demonstrating:  
+1. Access variable group values from Python. Note for sensitive variables ensure the variable group is linked to key vault. See https://learn.microsoft.com/en-us/azure/devops/pipelines/library/link-variable-groups-to-key-vaults?view=azure-devops
+2. Use of Service Principal Name (SPN) with a Secret credential flow, leveraging the ClientSecretCredential class. 
+3. Use the Fabric reset APIs to lookup the workspace ID based on workspace name
+4. Using debug log level
+"""
+# START-EXAMPLE
+
+# argparse is required to gracefully deal with the arguments
+import os,argparse, requests, ast
+from fabric_cicd import FabricWorkspace, publish_all_items, unpublish_all_orphan_items,change_log_level
+from azure.identity import ClientSecretCredential
+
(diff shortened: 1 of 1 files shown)


JacobKnightley:
Have you considered adding this as part of the library?  Accepting either workspace_id or workspace_name.  If name is provided, resolve the id at deployment.

hurtn:
> Have you considered adding this as part of the library? Accepting either workspace_id or workspace_name. If name is provided, resolve the id at deployment.

Yes happy to. Would you be okay for me to change the definition of the init method to be just workspace like this:

workspace: str | UUID

instead of workspace_id and then implement logic to determine if UUID or not in which case get the ID and set workspace_id variable before the validation function, or do you want me to put this logic in the validation function?

JacobKnightley:
Unfortunately we can't land any breaking changes.  Or at least need to be planful. We could add a warning to workspace_id that it will be deprecated in a future version.  For now making it a kwarg, but enforcing either workspace id or workspace must be provided.  And then yes use the validate function to properly handle and parse.

hurtn:
> Unfortunately we can't land any breaking changes. Or at least need to be planful. We could add a warning to workspace_id that it will be deprecated in a future version. For now making it a kwarg, but enforcing either workspace id or workspace must be provided. And then yes use the validate function to properly handle and parse.

The only way I think we can implement without a breaking change would be to allow the workspace_id to be either a UUID or Name (as many of the semantic link functions do already) so something like this...

def __init__(
    self,
    workspace_id: Union[str, UUID],
...

I do this sort of check in my code often like:

    if _is_valid_uuid(target_capacity):
        target_capacity_id = target_capacity.lower()
    else: 
        # fetch the capacity ID
        target_capacity_id = labs.resolve_capacity_id(target_capacity)

I think what you mean is that we need to keep the param name as workspace_id, which doesn't make it super clear that it can be either.

Let me know what you think otherwise I do agree it would be a larger (possibly breaking) change.

JacobKnightley:
@hurtn  this has been implemented direct in the library #200 .  Closing this now :)

PR #256, opened 2025-04-22
Fixes #255 Empty Folder Publishes

### LINKED ISSUE #255: Known Issue: Sub Folders Deploying to Root & Folder Already Exists
### Library Version

All

### What is the problem?

There is a known issue where folders are being deployed to the root of the workspace instead of to the proper structure.  In addition, because of this, then names are being reserved and failing to deploy.  

_No response_

_No response_

JacobKnightley:
This pull request refactors the `_refresh_repository_folders` method in `fabric_workspace.py` to simplify its logic and improve performance by leveraging `Path.rglob`. Additionally, it updates the method signature to return `None` instead of a dictionary, aligning with its new behavior.

### Method refactor for `_refresh_repository_folders`:
* Simplified folder traversal logic by using `Path.rglob` to collect folders containing `.platform` files and their subfolders, replacing the previous nested loops and multiple checks. This reduces complexity and improves maintainability.
* Updated the method signature to return `None` instead of a dictionary, as the method now directly assigns the processed folder hierarchy to `self.repository_folders`.

### DIFF
--- a/src/fabric_cicd/fabric_workspace.py
+++ b/src/fabric_cicd/fabric_workspace.py
@@ -534,7 +534,7 @@ class FabricWorkspace:
 
         self.deployed_folders = folder_hierarchy
 
-    def _refresh_repository_folders(self) -> dict:
+    def _refresh_repository_folders(self) -> None:
         """
         Converts the folder list payload into a structure of folder name and their ids,
         skipping empty folders or folders that only contain other empty folders.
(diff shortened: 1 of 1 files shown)


src/fabric_cicd/fabric_workspace.py
```diff
@@ -534,7 +534,7 @@ def get_full_path(folder: dict) -> str:
 
         self.deployed_folders = folder_hierarchy
 
-    def _refresh_repository_folders(self) -> dict:
+    def _refresh_repository_folders(self) -> None:
         """
         Converts the folder list payload into a structure of folder name and their ids,
```
jmuziki, line 539: the skipping logic has been changed - update docstrings to provide more detail on new approach.

jmuziki (approved):

PR #312, opened 2025-05-21
Resolves #273 Bug when determining orphan workspace folders to unpublish

### LINKED ISSUE #273: Unpublish shows warnings for workspace folders which only contain other folders
### Library Version

0.1.16

### What is the problem?

When the `unpublish_all_orphan_items` function is called for a workspace with a folder hierarchy which contains folders which contain no Fabric items and only contain other workspace folders, the following warning is logged for each such folder (some actual values replaced with placeholders):

```
[warn]   hh:mm:ss - Failed to unpublish folder ########-####-####-####-############.  Raw exception: Unhandled error occurred calling DELETE on 'https://api.powerbi.com/v1/workspaces/########-####-####-####-############/folders/########-####-####-####-############'. Message: The requested folder was not empty.. 
```

For example, if the workspace folder hierarchy looks like this:

```
📂 notebooks/
├─ 📂 sub_folder_1/
|  ├─ 📄 item_1
|  ├─ 📄 item_2
├─ 📂 sub_folder_2/
|  ├─ 📄 item_3
|  ├─ 📄 item_4
📂 pipelines/
├─ 📂 sub_folder_3/
|  ├─ 📄 item_5
├─ 📂 sub_folder_4/
|  ├─ 📄 item_6
```

There will be two warnings logged: one with the ID of the `notebooks` folder and one with the ID of the `pipelines` folder. These two folders don't directly contain any Fabric items (only folders which themselves contain items).

I assume there's a bug in the code that considers folders which only contain other folders as orphaned, and makes a call to the Fabric API to delete them. The Fabric API (correctly) determines they can't be deleted and returns an error, which is logged as a fabric-cicd warning.

Ideally fabric-cicd wouldn't consider any folder in the lineage path of any item as orphaned, and wouldn't attempt to delete them.

_No response_

shirasassoon:
This pull request includes updates to improve the handling of folder and item unpublishing in the `fabric_cicd` module. The most important changes involve optimizing the logic for determining orphaned folders and ensuring that folder metadata is refreshed during the unpublishing process.

### Improvements to folder unpublishing logic:
* [`src/fabric_cicd/fabric_workspace.py`](diffhunk://#diff-3bd0f70ed06c7fc7a0f77378aecc6fb108eeb5263161db3b22270f3de4df99c0L639-R659): Refactored the `_unpublish_folders` method to improve clarity and efficiency. The logic now uses a set to track folders containing items and their parent folders, ensuring that only orphaned folders are unregistered.

### Metadata refresh enhancements:
* [`src/fabric_cicd/publish.py`](diffhunk://#diff-a8555a82747092e8d1f2ee36f31c947b4e5a264ed6c6d8de48d8daf850372704R226): Added a call to `_refresh_deployed_folders` in the `unpublish_all_orphan_items` function to ensure folder metadata is up-to-date before unpublishing operations.

### DIFF
--- a/src/fabric_cicd/fabric_workspace.py
+++ b/src/fabric_cicd/fabric_workspace.py
@@ -636,21 +636,27 @@ class FabricWorkspace:
     def _unpublish_folders(self) -> None:
         """Unublishes all empty folders in workspace."""
         # Sort folders by the number of '/' in their paths (descending order)
-        sorted_folder_ids = [
-            self.deployed_folders[key]
-            for key in sorted(self.deployed_folders.keys(), key=lambda path: path.count("/"), reverse=True)
-        ]
-
+        sorted_folders = sorted(self.deployed_folders.keys(), key=lambda path: path.count("/"), reverse=True)
+        sorted_folder_ids = [self.deployed_folders[folder_path] for folder_path in sorted_folders]
         logger.info("Unpublishing Workspace Folders")
 
-        ## any folder that is not in folderid_dict is an orphaned folder
+        ## Any folder that does not contain an item or does not parent folders
+        ## containing an item in its hierarchy is an orphaned folder
 
-        # Get folders with items
-        deployed_folder_ids_with_items = []
-
-        for items in self.deployed_items.values():
-            for item in items.values():
-                deployed_folder_ids_with_items.append(item.folder_id)
+        # Create a set of folders that contain items, and therefore can't be deleted
+        deployed_folder_ids_with_items = {
+            item.folder_id for items in self.deployed_items.values() for item in items.values()
+        }
+        # Add parent folders of folders with items to the set
+        for folder_path in sorted_folders:
+            folder_id = self.deployed_folders[folder_path]
+            if folder_id in deployed_folder_ids_with_items:
+                path_parts = folder_path.split("/")
+                for i in range(1, len(path_parts)):
+                    parent_path = "/".join(path_parts[:i])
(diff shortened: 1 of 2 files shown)


src/fabric_cicd/fabric_workspace.py
```diff
@@ -636,21 +636,27 @@ def _publish_folders(self) -> None:
     def _unpublish_folders(self) -> None:
         """Unublishes all empty folders in workspace."""
```
jmuziki: typo change to "Unpublishes"

src/fabric_cicd/fabric_workspace.py
```diff
@@ -636,21 +636,27 @@ def _publish_folders(self) -> None:
...
 
-        for items in self.deployed_items.values():
-            for item in items.values():
-                deployed_folder_ids_with_items.append(item.folder_id)
+        # Create a set of folders that contain items, and therefore can't be deleted
+        deployed_folder_ids_with_items = {
+            item.folder_id for items in self.deployed_items.values() for item in items.values()
+        }
+        # Add parent folders of folders with items to the set
+        for folder_path in sorted_folders:
```
jmuziki: have you tested performance for this? if a folder has many nested subfolders this O(n^2) approach could be problematic

shirasassoon: @jmuziki that's a good call. Will work on improving the perf independent of number of folders

jmuziki (approved):

PR #424, opened 2025-07-17
 Closes #423 - Add support for 'all' option in item_types_in_scope

### LINKED ISSUE #423: Add support for 'all' option in item_type_in_scope parameter
### Description

Enhance the `item_type_in_scope` parameter to accept an `'all'` option that automatically includes all valid item types based on the authentication method (UPN vs non-UPN). This would simplify workspace initialization when users want to work with all available item types without manually specifying each one.

### Current behavior

Users must manually specify all item types:
```python
item_type_in_scope=["Environment", "Notebook", "DataPipeline", "Lakehouse", "VariableLibrary", "CopyJob", "Eventhouse", "KQLDatabase", "KQLQueryset", "Reflex", "Eventstream", "Warehouse", "SQLDatabase", "KQLDashboard", "Dataflow", "GraphQLApi", "Report", "SemanticModel", "MirroredDatabase"]
```

### Proposed behavior

Allow users to use a simple `'all'` option:
```python
item_type_in_scope=["all"]
```

This feature would automatically resolve to the appropriate set of item types from `constants.ACCEPTED_ITEM_TYPES_UPN` or `constants.ACCEPTED_ITEM_TYPES_NON_UPN` based on the authentication method being used.

### Example Usage

```python
from fabric_cicd import FabricWorkspace

# Instead of this:
workspace = FabricWorkspace(
    workspace_id="your-workspace-id",
    repository_directory="/path/to/repo",
    item_type_in_scope=["Environment", "Notebook", "DataPipeline", "Lakehouse", ...]
)

# Users can now do this:
workspace = FabricWorkspace(
    workspace_id="your-workspace-id",
    repository_directory="/path/to/repo",
    item_type_in_scope=["all"]
)
```

### Additional context

_No response_

stanlocht:
## Description

This PR implements the `'all'` option for the `item_type_in_scope` parameter, allowing users to automatically include all valid item types without manually specifying each one.

**Modified `validate_item_type_in_scope()` function** in `src/fabric_cicd/_common/_validate_input.py`:
  - Added logic to detect when `item_type_in_scope` contains only `["all"]` (case-insensitive)
  - Returns appropriate item types based on authentication method (`upn_auth` parameter)
  - Maintains backward compatibility with existing specific item type lists
  - Added informative logging when 'all' option is used

## Linked Issue (REQUIRED)
 Closes #423


stanlocht:

JacobKnightley:
I would suggest a slightly different approach.  We can make item_type_in_scope to be optional.  And if not provided, then we set item_type_in_scope to the accepted_item_types.  Although, the question is, would you really want us to attempt to deploy all item types?  Or should we traverse the directory to identify what item types are in your repository, and deploy whatever is there?

stanlocht:
@JacobKnightley Yes, I agree that publishing all items as default behaviour would be nice. In my experience, most of the time I just want to publish all items found in the repo. The downstream logic of not attempting all item types, but only those found in the repository seems better, but to implement that might be out of scope for this PR? 

What would you suggest as an approach for now? We could make the ` item_type_in_scope` param optional for now, to later add functionality to find all item types present in the repo?

stanlocht:
@JacobKnightley Do you think we could pick up this PR? What approach would you suggest?

JacobKnightley:
I'm going to merge this, and then I will build on top of it to enable the default behaviors :)

JacobKnightley (approved):

cavassiGE:
@JacobKnightley would it be possible to do something similar to _unpublish_item or unpublish_all_orphan_items? So you could use "all" as an argument and easily delete every item in the workspace. Or insert what item type you want to delete. Can for example be useful when you want to copy an item, or want the same item but new logical ID.

PR #453, opened 2025-08-05
Update _fabric_endpoint.py to handle JSONDecodeErrors from the response handler

### LINKED ISSUE #452: Custom Endpoint hadling via Invoke
### Library Version

0.1.23

### What is the problem?

When using FabricWorkspace.endpoint.invoke on the jobs/instances endpoint to run notebooks the _response handler method causes the function call error as response_json() cases a JSONDecodeError as there is no content in the response.

To reproduce call "https://api.fabric.microsoft.com/v1/workspaces/<workspace_id>/items/<item_id>/jobs/instances?jobType=runNotebook" on any notebook item.

I believe the solution is to mimic the body definition in the invoke methods return statement to avoid JSONDecodeErrors.

_No response_

Maki-DS:
## Description

This response handler would error if you tried to execute a run notebook job as response.json() gives a JSONDecodeError. I have only updated the response_json = response.json() to mimic that of the return statement of the invoke method.

## Linked Issue (REQUIRED)
https://github.com/microsoft/fabric-cicd/issues/452

### DIFF
--- a/src/fabric_cicd/_common/_fabric_endpoint.py
+++ b/src/fabric_cicd/_common/_fabric_endpoint.py
@@ -180,7 +180,9 @@ def _handle_response(
     # Handle long-running operations
     # https://learn.microsoft.com/en-us/rest/api/fabric/core/long-running-operations/get-operation-result
     if (response.status_code == 200 and long_running) or response.status_code == 202:
+        print(response.headers)
         url = response.headers.get("Location")
+        print(url)
         method = "GET"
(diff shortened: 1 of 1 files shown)


Maki-DS:
Please ignore this for now. I have noticed that this works however it will try to get a new Location for jobs/instances. Feels to me like the retry logic is for a specific api call rather than the jobs scenario. I will keep looking and try to make it work (maybe add a named param?).

PR #457, opened 2025-08-05
Allow _handle_response method to handle more long running events

### LINKED ISSUE #452: Custom Endpoint hadling via Invoke

Maki-DS:
## Description
Allows _handle_responses method in _fabric_endpoint.py to handle jobs/instance post calls (wont error on response_json = response.json() and checks for the "Completed" status.

## Linked Issue (REQUIRED)
https://github.com/microsoft/fabric-cicd/issues/452

### DIFF
--- a/src/fabric_cicd/_common/_fabric_endpoint.py
+++ b/src/fabric_cicd/_common/_fabric_endpoint.py
@@ -190,7 +190,7 @@ def _handle_response(
             if status in ["Succeeded", "Completed"]:
                 long_running = False
                 # If location not included in operation success call, no body is expected to be returned
-                exit_loop = url is None
+                exit_loop = location is None
 
             elif status == "Failed":
                 response_error = response_json["error"]


Maki-DS:

JacobKnightley (changes requested):
Unit tests failed.  Please let's chat in the Issue before making further edits as I'm not confident fabric-cicd is the right tool for your scenario.

shirasassoon:
@Maki-DS Closing due to aging.

PR #584, opened 2025-10-03
Fixes #583 - Workspace Name Not Extracted When Using Non-ID Attribute

### LINKED ISSUE #583: Support other attributes in parameterization cross-workspace items variable
### Library Version

0.1.29

### What is the problem?

I am trying to get the SQL endpoint of a warehouse in a report of mine. Instead of getting the SQL endpoint replaced in the report, I get the following deployment message:

```
Workspace ID could not be resolved from workspace name: TST_MyWorkspace.$items.Warehouse.WH_MyWarehouse.$sqlendpoint.
```

parameter.yml
```
- find_value: "mydevwarehouse.datawarehouse.fabric.microsoft.com"
  replace_value:
    TST: "$workspace.TST_MyWorkspace.$items.Warehouse.WH_MyWarehouse.$sqlendpoint"
```

I believe this is related to [this line](https://github.com/microsoft/fabric-cicd/blob/3ba50f2d52a1d3f11b308e3c51e8e5cb8a8fdc37/src/fabric_cicd/_parameter/_utils.py#L110), where the workspace assumed to be the full string after `$workspace.` only when the string ends in `.id$` but the string could end in `.$sqlendpoint` or `.$queryserviceuri`

_No response_

camronbute-lantern:
This pull request adds a test that fails in version v0.1.29 when testing that the workspace name is correctly extracted when looking up a non-`id` attribute (such as `sqlendpoint`) for an item in a different workspace.

The fix was to reference the `ITEM_ATTR_LOOKUP` constant when deciding which format algorithm to use for workspace name extraction.

### DIFF
--- a/src/fabric_cicd/_parameter/_utils.py
+++ b/src/fabric_cicd/_parameter/_utils.py
@@ -107,7 +107,9 @@ def _extract_workspace_id(workspace_obj: FabricWorkspace, replace_value: str) ->
         var_string = replace_value.removeprefix("$workspace.")
 
         # Check for pattern: $workspace.name.$items.type.name.$id
-        if "$items." in var_string and var_string.endswith(".$id"):
+        if "$items." in var_string and var_string.endswith(
+            tuple(f"{prefix}{suffix}" for suffix in constants.ITEM_ATTR_LOOKUP for prefix in [".$", "."])
+        ):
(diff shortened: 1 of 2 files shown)


camronbute-lantern:
Duplicate of #585

PR #594, opened 2025-10-14
Fixes #593 - Fixes capacity requirement check to work regardless of item type order

### LINKED ISSUE #593: Error in how deployment checks if a Fabric capacity is needed
### Library Version

0.1.28

### What is the problem?

There is an error in how the library checks if the items_in_scope would require a fabric capacity for successful deployment.
If i use `item_type_in_scope = ["Report", "SemanticModel"]`, the code fails in the preflight check. If I change this to `item_type_in_scope = ["SemanticModel","Report"]` then the deployment works.

It appears that the logic for checking what the non-capacity items are is flawed.

It should not matter what order the item types are defined :)

The line 111 in publish.py should be changed to 
`and not set(fabric_workspace_obj.item_type_in_scope).issubset(set(constants.NO_ASSIGNED_CAPACITY_REQUIRED))`. 
And at the same time the constants.NO_ASSIGNED_CAPACITY_REQUIRED should be reduced to just `["SemanticModel", "Report"]` (instead of `[["SemanticModel", "Report"], ["SemanticModel"], ["Report"]]`)

_No response_

celindho:
## Description

This fixes how the code validates whether a Fabric capacity is needed or not. The original code required the "item types in scope" to be written in a specific order, this changes the validation to work regardless of item type order.

## Linked Issue (REQUIRED)

- Fixes #593

### DIFF
--- a/src/fabric_cicd/constants.py
+++ b/src/fabric_cicd/constants.py
@@ -41,7 +41,7 @@ ACCEPTED_ITEM_TYPES = (
 SHELL_ONLY_PUBLISH = ["Environment", "Lakehouse", "Warehouse", "SQLDatabase"]
 
 # Items that do not require assigned capacity
-NO_ASSIGNED_CAPACITY_REQUIRED = [["SemanticModel", "Report"], ["SemanticModel"], ["Report"]]
+NO_ASSIGNED_CAPACITY_REQUIRED = ["SemanticModel", "Report"]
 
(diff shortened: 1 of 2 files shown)


celindho:
@microsoft-github-policy-service agree company="Fellowmind"

JacobKnightley:
Good catch!  @celindho

PR #682, opened 2025-12-12
Fixes #674: Parameterizing Multiple Connections in a Single Semantic Model

### LINKED ISSUE #674: Parameterizing Multiple Connections in a Single Semantic Model Using semantic_model_binding Throwing error.
We are using the semantic_model_binding parameterization feature as documented, and we observed that when the same semantic model is listed more than once with different connection IDs, pipeline is failing.

Image

Image

Could you please clarify whether semantic_model_binding is intentionally designed to support only one connection per semantic model?

If a semantic model legitimately contains multiple data source connections, what is the recommended or supported approach to parameterize and bind multiple connections for a single semantic model during deployment using fabric-cicd?

falas6ene81:
This pull request updates parameter validation logic to improve maintainability and ensure all relevant parameters are checked consistently. The main changes involve centralizing the list of parameters to validate and updating validation functions to use this centralized list.

Parameter validation improvements:

* Added a new constant `PARAM_NAMES` in `constants.py` to define all parameter names that require validation, including `"semantic_model_binding"`.
* Updated `is_valid_structure` in `_utils.py` to use `PARAM_NAMES` from `constants.py` instead of a hardcoded list, ensuring all relevant parameters are checked consistently.

Code simplification:

* Removed the redundant validation check for `"semantic_model_binding"` in `_validate_required_values` in `_parameter.py`, since it is now handled by the centralized parameter list.

### DIFF
--- a/activate.sh
+++ b/activate.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+#
+#
+#       Script to check and install required Python packages, 
+#       add directories to PATH, and activate a virtual environment.
+#
+# ---------------------------------------------------------------------------------------
+#
+set -e
+
+PACKAGES=""
+if ! command -v python &> /dev/null; then PACKAGES="python3"; fi
+if ! command -v pip &> /dev/null; then PACKAGES="${PACKAGES:+$PACKAGES }python3-pip"; fi
+if ! command -v ruff &> /dev/null; then PACKAGES="${PACKAGES:+$PACKAGES }ruff"; fi
+if [ -n "$PACKAGES" ]; then
+    sudo apt-get update > /dev/null 2>&1
+    sudo DEBIAN_FRONTEND=noninteractive apt-get install -y $PACKAGES > /dev/null 2>&1
+fi
+command -v uv &> /dev/null || curl -LsSf https://astral.sh/uv/install.sh | sh
+
+[[ ":$PATH:" != *":$HOME/.local/bin:"* ]] && export PATH="$PATH:$HOME/.local/bin"
+[[ ":$PATH:" != *":$HOME/.cargo/bin:"* ]] && export PATH="$PATH:$HOME/.cargo/bin"
+
+uv sync --python 3.11
+[ -f .venv/bin/activate ] && source .venv/bin/activate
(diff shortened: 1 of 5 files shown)


src/fabric_cicd/_parameter/_parameter.py
```diff
@@ -550,11 +550,6 @@ def _validate_required_values(self, param_name: str, param_dict: dict) -> tuple[
             if not is_valid:
                 return False, msg
 
-        if param_name == "semantic_model_binding":
-            is_valid, msg = self._validate_semantic_model_name()
-            if not is_valid:
-                return False, msg
-
         if param_name == "key_value_replace":
```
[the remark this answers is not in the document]
falas6ene81: users can bind the same model to multiple connections.

src/fabric_cicd/constants.py
```diff
@@ -74,6 +74,9 @@
 
 # Parameter file configs
 PARAMETER_FILE_NAME = "parameter.yml"
+# Parameters to validate
+PARAM_NAMES = ["find_replace", "key_value_replace", "spark_pool", "semantic_model_binding"]
```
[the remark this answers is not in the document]
falas6ene81: ok not today

src/fabric_cicd/_parameter/_parameter.py
shirasassoon: If we're removing the check, shouldn't we remove the validation logic entirely or is this still being used elsewhere?

src/fabric_cicd/_parameter/_utils.py
```diff
@@ -518,11 +518,8 @@ def is_valid_structure(param_dict: dict, param_name: Optional[str] = None) -> bo
     if param_name:
         return _check_parameter_structure(param_dict.get(param_name))
 
-    # Parameters to validate
-    param_names = ["find_replace", "key_value_replace", "spark_pool"]
-
     # Get only parameters that exist in param_dict
-    existing_params = [name for name in param_names if name in param_dict]
+    existing_params = [name for name in constants.PARAM_NAMES if name in param_dict]
```
shirasassoon, line 522: Let's ensure the structure validation passes with this update.

src/fabric_cicd/_parameter/_parameter.py
falas6ene81: I left it in case you it's needed later. We might run into a scenario in which we have to check for duplicate semantic model names, but, happy to remove it.

PR #697, opened 2025-12-31
Closes #115 - Support Spark Job Definitions

### LINKED ISSUE #115: Support for Spark Job Defintions
### What is the feature?

APIs are publicly available to onboard Spark Job Definitions. However, during the initial investigation we found that the general process does not align to how other Fabric artifacts can be deployed via the APIs. Namely that users must first deploy a shell and then upload files/libraries to Onelake separately before running an additional flush. We are chatting internally on ways to improve this before onboarding.

### Additional context

_No response_

mwc360:
## Description

Adds support for Spark Job Definitions using the v2 API. The v2 API supports creation of SJDs with dependent (Main/Lib) files that are base64 encoded strings.

Per the [API Docs](https://learn.microsoft.com/en-us/fabric/data-engineering/spark-job-definition-api-v2#spark-job-definition-v2-format-overview), only .scala and .py files are supported for encoded files in the SJD definition. To include other types an ABFSS path to a file can be provided instead.


src/fabric_cicd/publish.py
```diff
@@ -407,6 +410,7 @@ def unpublish_all_orphan_items(
         "CopyJob",
         "Report",
         "SemanticModel",
+        "SparkJobDefinition",
```
shirasassoon: @mwc360 For consistency let's place SparkJobDefinition item in exact reverse order. It should be placed between Notebook and Eventhouse.

docs/how_to/item_types.md
```diff
@@ -145,6 +145,12 @@
     -   Semantic Models connected to sources within the same workspace may or may not be re-pointed; it is best to test this before taking a dependency. Use the `find_replace` section of the `parameter.yml` file as needed.
 -   **Initial deployment** requires manual configuration of the connection after deployment.
 
+## Spark Job Definitions
+
+-   **Parameterization:**
+    -   Spark Job Definitions attached to lakehouses always point to the original lakehouse unless parameterized in the `find_replace` section of the `parameter.yml` file.
```
shirasassoon, line 151: Let's mention Environment reference as well. Also, format files this way for clearer readability - `.py`, `.scala` etc.

src/fabric_cicd/_items/_sparkjobdefinition.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Publishes all sparkjobdefinition items from the repository.
+
+    Args:
+        fabric_workspace_obj: The FabricWorkspace object containing the items to be published.
+    """
+    item_type = "SparkJobDefinition"
+
+    for item_name in fabric_workspace_obj.repository_items.get(item_type, {}):
+        fabric_workspace_obj._publish_item(item_name=item_name, item_type=item_type, format="SparkJobDefinitionV2")
```
shirasassoon: Consider renaming the parameter to `api_format` for clearer understanding and adding a brief comment above  the line of code or add a comment by the logic in `FabricWorkspace`

src/fabric_cicd/fabric_workspace.py
```diff
@@ -620,7 +620,10 @@ def _publish_item(
 
                     item_payload.append(file.base64_payload)
 
-            definition_body = {"definition": {"parts": item_payload}}
+            if kwargs.get("format"):
```
shirasassoon: Add a comment to briefly describe the logic

shirasassoon (changes requested):
@mwc360 Thanks for your contribution to fabric-cicd! Please ensure all changes are tested and work as expected. 

Please review some minor comments.

docs/how_to/item_types.md
mwc360: Updated. Is there a reason why auto replacement happens for an EnvironmentID but not for a LakehouseID? Looks like there's a function which swaps logical IDs with the actual deployed GUID. Seems odd for it to not be consistent by Item type.

shirasassoon: @mwc360 It seems you are pointing out that environment id is taken care of automatically versus lakehouse id isn't that would mean the source control file uses logical id for environment item reference and the item id for the lakehouse reference. Can you please confirm this is the case? If it is, you can optionally leave the note on environment (whatever you think it clear for end users).

shirasassoon: To answer your question, Logical ID replacement gets applied to _any_ item type as long as it is referenced by the logical id and not the item id.

mwc360:
This is just my observation. From the code, I don’t understand why there’s a diff between EnvironmentID and LakehouseID. Can you explain this. I don’t feel comfortable expanding the docs if I don’t understand why it works that way. Thx!

Miles Cole
Principal Program Manager
Fabric CAT | Spark
***@***.******@***.***>

[signatureImage]
________________________________

docs/how_to/item_types.md
mwc360: @shirasassoon Ok, I understand what's going on now. Some item types (Lakehouse, Warehouse, etc.) generate a `logicalId` in GIT that is different than the source workspace `ItemId`. This breaks automatic ID replacement and therefore requires referencing Notebooks to use param replacement to reference the deployed Lakehouse. Others like Environments, Notebooks, etc. get a `logicalId` that is the same as the `ItemId` and therefore the `fabric-cicd` auto replacement works beautifully.

So my doc commit is still accurate. I'd prefer not to update doc references on this topic other places for now. I think it would be best to connect with the platform team to get a comprehensive list of Items that get new logicalIds with GIT sync first and then make a focused doc update. PR is good to go unless there's any other comments.

devtools/debug_local.py
```diff
@@ -48,6 +48,7 @@
     "KQLQueryset",
     "Reflex",
     "Eventstream",
+    "SparkJobDefinition",
```
shirasassoon, line 51: @mwc360 Can you confirm if a user runs a deployment with the sample SJD item you provided in this PR, the deployment will work? I see it is referencing Lakehouse and Environment items, it may result in a deployment error if those don't exist.  Ideally sample items should work if a user chooses to deploy it.

mwc360: Yes, I ran this debug_local.py and it works. I am also separately using a build of this WHL to deploy solutions that include SJDs.

shirasassoon: @mwc360 That makes sense, however, where do the referenced items sit? I assume they were created by you so the deployment would work, but would it work if I or another user would run this.

mwc360: confirming as discussed separately, a SJD was added to the `sample/workspace` and tested locally.

PR #722, opened 2026-01-09
perf: Parallelize Every Item #721

### LINKED ISSUE #719: Is there a reason fabric-cicd does not deploy objects parallely?
### What is the feature?

For example, if I have 100 lakehouses with 100 tables each, each of these deploys serially, it's `O(N)`:

E.g. line 28 deploys lakehouses serially:

https://github.com/microsoft/fabric-cicd/blob/55aad3330fc4fa81a31fa5652e660fdfafb0247e/src/fabric_cicd/_items/_lakehouse.py#L28

And then per lakehouse, line 55 deploys per shortcut serially:

https://github.com/microsoft/fabric-cicd/blob/55aad3330fc4fa81a31fa5652e660fdfafb0247e/src/fabric_cicd/_items/_lakehouse.py#L55

At larger scale setups, having parallelization significantly reduces the time to deploy

With `Environments` for Spark, this gets even worse.

I understand Fabric can throttle on too many API calls, but we can:

1. Batch them up where possible (e.g. Shortcuts has a batch API)
2. Retry on throttle

---

If there's no technical reason to have serial, would you be open to having parallel deployment logic? E.g. even if it's under a feature flag, I'm happy to try it out.

### Additional context

Our deployments takes 2 hours and times out because we have 100s of Lakehouses, Shortcuts and Environments 😪

Image

mdrakiburrahman:
> 😧
> image
>
> ⚠️ Don't panic!
> 
> The vast majority of these diffs are in `http_trace.json`, which contains a snapshot of the Fabric REST API requests and responses (no secrets committed) for a Mock API integration test described below 🙂
>
> 
> image

## Description

This PR adds parallelism for every item by default and allows each item to override the behavior for those that must be serial.

This results in significant reduction in deployment execution time.

Also, since more parallel API calls means increased probability of throttling, rather than the current static `5` retries per `invoke`, we opt in for a maximum duration cap instead. This along with the linear back-off drastically increases the probability of success of a cicd run, despite throttling.

The PR also centralizes all:

* Feature flags
* Env-vars
* Execution sequence

Etc. so one can override behavior without changing the public facing API (e.g. in tests).

Centralizing also helps reason about the configurable knobs as an outsider.

## Linked Issue (REQUIRED)

Resolves #719, #726

## Before and After for ~20 Lakehouses and ~200 tables Average 10 runs

image

About a **100%+** improvement in execution time.

The more items one deploys of a particular type, the more this benefits.

### Mock Fabric REST API

As this PR contains a significant amount of structural changes, mocking is not sufficient and a black box test harness is required.

The PR includes a Mock Fabric REST API Server, that uses a snapshot of HTTP calls. This allows us to test the CICD project against a blackbox:

test-harness

### Real Workspace Test

I deployed this branch's build to 9 of our Fabric Workspaces to ensure no regressions:

image

### DIFF
--- a/tests/fixtures/README.md
+++ b/tests/fixtures/README.md
@@ -2,11 +2,16 @@
 
 ## `mock_fabric_server`: A mock Fabric REST API
 
-![Test Harness](.imgs/test-harness.png)
-
 This is a mock REST API Server that mimics `https://api.powerbi.com`.
 The idea is, to exercise the public facing `fabric_cicd` API E2E rapidly.
 The mock server loads an `http_trace.json` file to dictate it's behavior.
+
+![Test Harness](.imgs/test-harness.png)
+
+1. Add new workloads into the codebase
+2. Capture REST calls from Fabric using `debug_trace_publish_all.py`
+3. Move `http_trace.json` into fixture
+4. Enjoy rapid test coverage
 
 ### Capturing HTTP Trace for new resource types
 


src/fabric_cicd/_items/_lakehouse.py
```diff
@@ -44,19 +46,25 @@ def publish_lakehouses(fabric_workspace_obj: FabricWorkspace) -> None:
...
         if item.skip_publish:
-            continue
+            return
 
         check_sqlendpoint_provision_status(fabric_workspace_obj, item)
 
         logger.info(f"{constants.INDENT}Published")
 
+    with ThreadPoolExecutor() as executor:
+        list(executor.map(_publish, fabric_workspace_obj.repository_items.get(item_type, {}).items()))
```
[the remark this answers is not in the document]
mdrakiburrahman: That's the goal. Fail fast was the previous behavior when previous runs were sequential.

I did not want to refactor the code too much. `executor.map` is the cleanest approach because it effectively parallelizes the for loop.

src/fabric_cicd/_items/_lakehouse.py
```diff
@@ -170,10 +179,13 @@ def publish_shortcuts(fabric_workspace_obj: FabricWorkspace, item_obj: Item, sho
                     f"Failed to publish '{shortcut['name']}'. This usually happens when the lakehouse containing the source for this shortcut is published as a shell and has no data yet."
                 )
                 logger.info("The publish process will continue with the other items.")
-                continue
+                return
             msg = f"Failed to publish '{shortcut['name']}' for lakehouse {item_obj.name}"
             raise FailedPublishedItemStatusError(msg, logger) from e
 
+    with ThreadPoolExecutor() as executor:
+        list(executor.map(_publish_one, shortcut_dict.values()))
```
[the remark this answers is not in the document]
mdrakiburrahman: Same as above:

https://github.com/microsoft/fabric-cicd/pull/722#discussion_r2677866654

src/fabric_cicd/_items/_lakehouse.py
```diff
@@ -154,7 +162,8 @@ def publish_shortcuts(fabric_workspace_obj: FabricWorkspace, item_obj: Item, sho
         item_obj: The item object to publish shortcuts for
         shortcut_dict: The dict of shortcuts to publish
     """
-    for shortcut in shortcut_dict.values():
+
+    def _publish_one(shortcut: dict) -> None:
         shortcut = replace_default_lakehouse_id(shortcut, item_obj)
```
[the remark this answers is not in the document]
mdrakiburrahman: This was exactly the previous behavior, I do not want to change it. The diff here just wraps the existing code in a function.

src/fabric_cicd/_items/_environment.py
```diff
@@ -51,9 +53,12 @@ def publish_environments(fabric_workspace_obj: FabricWorkspace) -> None:
             shell_only_publish=is_shell_only,
         )
         if item.skip_publish:
-            continue
+            return
         _publish_environment_metadata(fabric_workspace_obj, item_name)
 
+    with ThreadPoolExecutor() as executor:
+        list(executor.map(_publish, fabric_workspace_obj.repository_items.get(item_type, {}).items()))
```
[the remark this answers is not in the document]
mdrakiburrahman: This is untrue, this code is tested. I'm not testing for parallel behavior, true, but we shouldn't be testing implementation details - the public interface `publish_lakehouses` is tested.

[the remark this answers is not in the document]
mdrakiburrahman: Same as above:

https://github.com/microsoft/fabric-cicd/pull/722#discussion_r2677866654

src/fabric_cicd/_items/_lakehouse.py
```diff
@@ -44,19 +46,25 @@ def publish_lakehouses(fabric_workspace_obj: FabricWorkspace) -> None:
...
+        def _process(item_obj: Item) -> None:
             # Check if the item is published to avoid any post publish actions
-            if item_obj.skip_publish:
-                continue
-            process_shortcuts(fabric_workspace_obj, item_obj)
+            if not item_obj.skip_publish:
+                process_shortcuts(fabric_workspace_obj, item_obj)
+
+        with ThreadPoolExecutor() as executor:
+            list(executor.map(_process, fabric_workspace_obj.repository_items.get(item_type, {}).values()))
```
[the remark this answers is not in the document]
mdrakiburrahman: Same as above:

https://github.com/microsoft/fabric-cicd/pull/722#discussion_r2677866654

src/fabric_cicd/_items/_lakehouse.py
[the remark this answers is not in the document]
mdrakiburrahman: Same as above:

https://github.com/microsoft/fabric-cicd/pull/722#discussion_r2677876261

tests/test__fabric_endpoint.py
```diff
@@ -401,8 +409,10 @@ def test_handle_response_longrunning_exception(exception_match, response_json):
             {"Content-Type": "application/json"},
             {"message": "Internal Server Error"},
             "Unhandled error occurred calling GET on 'http://example.com'. Message: Internal Server Error",
+            None,
+            None,
         ),
-        (429, 5, True, {"Retry-After": "10"}, {}, r"Maximum retry attempts \(5\) exceeded."),
+        (429, 5, True, {"Retry-After": "10"}, {}, r"Maximum execution duration \(0 seconds\) exceeded", 0, 0.0),
```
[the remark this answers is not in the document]
mdrakiburrahman: I am trying to preserve the existing format as much as possible - previously, regex was not being used.

src/fabric_cicd/_items/_lakehouse.py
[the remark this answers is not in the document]
mdrakiburrahman: Same as above:

https://github.com/microsoft/fabric-cicd/pull/722#discussion_r2677876261

src/fabric_cicd/_common/_fabric_endpoint.py
```diff
@@ -243,10 +251,23 @@ def _handle_response(
...
             attempt=iteration_count,
             base_delay=10,
-            max_retries=5,
+            max_duration=max_duration,
+            start_time=start_time,
             response_retry_after=retry_after,
             prepend_message="API is throttled.",
         )
+    
+    # Handle internal server errors via retry,
```
mdrakiburrahman, line 276: This was really interesting behavior from the Control Plane Server. At random times, it flip flops between 429 and 500 - so I added this to make the client resilient, since the Server is behaving non-deterministically.

mdrakiburrahman: Note that this only reproduces when you have a very large number of lakehouse/shortcut tables. For smaller Workspaces all is well.

src/fabric_cicd/_common/_fabric_endpoint.py
```diff
@@ -334,7 +358,8 @@ def handle_retry(
         )
         time.sleep(delay)
     else:
-        msg = f"Maximum retry attempts ({max_retries}) exceeded."
+        elapsed = time.time() - start_time if start_time is not None else 0
+        msg = f"Maximum execution duration ({max_duration} seconds) exceeded after {elapsed:.1f} seconds."
```
shirasassoon, line 382: Maybe add a brief comment to the modified code in this section so it's clear to other developers on what is being done

src/fabric_cicd/_common/_fabric_endpoint.py
```diff
@@ -46,6 +46,7 @@ def invoke(
         body: str = "{}",
         files: Optional[dict] = None,
         poll_long_running: bool = True,
+        max_duration: int = 300,
```
shirasassoon, line 59: I'm ok with these changes as long as they have been tested thoroughly and reliability has been maintained. Ensure unit tests coverage includes the changes.

src/fabric_cicd/_items/_environment.py
```diff
@@ -33,10 +34,11 @@ def publish_environments(fabric_workspace_obj: FabricWorkspace) -> None:
     )
 
     # Check for ongoing publish
+    item_type = "Environment"
     check_environment_publish_state(fabric_workspace_obj, True)
 
-    item_type = "Environment"
-    for item_name, item in fabric_workspace_obj.repository_items.get(item_type, {}).items():
+    def _publish(args: tuple) -> None:
```
shirasassoon: I think parallelization makes sense, I would recommend centralizing the logic so it can be available for all supported item types. I would add the common parallelization logic to a file under the `_common` module and then reference the functionality across all item type files. 

Obviously determine if the new format works well end to end and include unit test coverage.

src/fabric_cicd/_items/_lakehouse.py
```diff
@@ -25,7 +26,8 @@ def publish_lakehouses(fabric_workspace_obj: FabricWorkspace) -> None:
     """
     item_type = "Lakehouse"
 
-    for item_name, item in fabric_workspace_obj.repository_items.get(item_type, {}).items():
+    def _publish(args: tuple) -> None:
```
shirasassoon: Before it was clean and easy to follow the item publish flow. I think if this logic sits in a central location it will maintain a flow similar to the original.

shirasassoon (commented):
@mdrakiburrahman If you have confidence in the performance and approach I don't think it's necessary to hide under a feature flag. I would like to see the parallelization logic centralized and applied to other item types (notebook is a good one too). Once you come up with a working solution, we should include a new test file with integration and unit tests to ensure the new parallelization logic is covered and works as expected.

src/fabric_cicd/_common/_fabric_endpoint.py
mdrakiburrahman: Done - the structure of the functions have been moved into classes, each has docstrings.

src/fabric_cicd/_items/_environment.py
mdrakiburrahman: Done, created constructs in `_base_publisher`.

By default, most items are parallel, except items that have inherent dependencies (like DataFlow and Pipelines), those override to Serial execution.

src/fabric_cicd/_items/_lakehouse.py
mdrakiburrahman: Done, I moved the logic into `publish_one` interfaced class.

src/fabric_cicd/_common/_fabric_endpoint.py
mdrakiburrahman: Done, added a Integration test harness with a Mock Fabric REST API. Please see `README` under PyTest fixtures.

src/fabric_cicd/_items/_environment.py
```diff
@@ -280,3 +245,34 @@ def _convert_environment_compute_to_camel(fabric_workspace_obj: FabricWorkspace,
...
+            skip_publish_logging=True,
+            shell_only_publish=is_shell_only,
+        )
+        if item.skip_publish:
+            return
+        _publish_environment_metadata(self.fabric_workspace_obj, item_name)
+
+    def pre_publish_all(self) -> None:
+        """Check environment publish state before publishing."""
+        logger.warning("The underlying legacy Microsoft Fabric Environment APIs will be deprecated by March 1, 2026.")
```
JacobKnightley, line 274: nit.  this isn't needed anymore. not related to this PR, but still.

mdrakiburrahman: Got it, will remove

mdrakiburrahman:
Spoke to @shirasassoon and @JacobKnightley  - it's best we split the PR up in 2, test harness and the class/parallel factoring.

1 is here: #739 

When 1 merges, I'll rebase this branch fresh, and open a PR 2.

PR #761, opened 2026-01-26
perf: Parallelize all items by central encapsulation into base classes

### LINKED ISSUE #719: Is there a reason fabric-cicd does not deploy objects parallely?

mdrakiburrahman:
## Description

Parallelizes deployments within a given item type.

This required changes to all `item` files, in order to push as much logic into the base classes as possible.
Item specific logic remains in the implementations via override.

> See `src/fabric_cicd/_items/_base_publisher.py` for the bulk of the logic

We do not parallelize _across_ item types, as the dependencies are non-deterministic at this time.
 
## Linked Issue

- Closes #719

## Test

Tested `devtools/debug_trace_deployment.py`:

image

Tested across 3 workspaces:

image

### DIFF
--- a/src/fabric_cicd/_items/__init__.py
+++ b/src/fabric_cicd/_items/__init__.py
@@ -1,7 +1,8 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-from fabric_cicd._items._base_publisher import ItemPublisher, ParallelConfig, PublishError
+from fabric_cicd._common._exceptions import PublishError
+from fabric_cicd._items._base_publisher import ItemPublisher, ParallelConfig
 
 __all__ = [
     "ItemPublisher",
--- a/src/fabric_cicd/_common/_exceptions.py
+++ b/src/fabric_cicd/_common/_exceptions.py
@@ -56,3 +56,24 @@ class FailedPublishedItemStatusError(BaseCustomError):
 
 class MissingFileError(BaseCustomError):
     pass
+
+
+class PublishError(BaseCustomError):
+    """Exception raised when one or more publish operations fail.
+
+    Attributes:
+        errors: List of (item_name, exception) tuples for all failed items.
+    """
+
+    def __init__(self, errors: list[tuple[str, Exception]], logger: Logger) -> None:
+        """Initialize with a list of (item_name, exception) tuples."""
+        self.errors = errors
+        failed_names = [name for name, _ in errors]
+        message = f"Failed to publish {len(errors)} item(s): {failed_names}"
+
+        additional_info_parts = []
+        for item_name, exc in errors:
+            additional_info_parts.append(f"\n--- {item_name} ---\n{exc!s}")
+        additional_info = "\n".join(additional_info_parts) if additional_info_parts else None
+
+        super().__init__(message, logger, additional_info)
--- a/src/fabric_cicd/_items/_base_publisher.py
+++ b/src/fabric_cicd/_items/_base_publisher.py
@@ -9,6 +9,7 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass
 from typing import Callable, Optional
 
+from fabric_cicd._common._exceptions import PublishError
 from fabric_cicd._common._item import Item
 from fabric_cicd.constants import ItemType
 from fabric_cicd.fabric_workspace import FabricWorkspace
@@ -35,20 +36,6 @@ class ParallelConfig:
     enabled: bool = True
     max_workers: Optional[int] = None
     ordered_items_func: Optional[Callable[["ItemPublisher"], list[str]]] = None
-
-
-class PublishError(Exception):
-    """Exception raised when one or more publish operations fail.
-
-    Attributes:
-        errors: List of (item_name, exception) tuples for all failed items.
-    """
-
-    def __init__(self, errors: list[tuple[str, Exception]]) -> None:
-        """Initialize with a list of (item_name, exception) tuples."""
-        self.errors = errors
-        failed_names = [name for name, _ in errors]
-        super().__init__(f"Failed to publish {len(errors)} item(s): {failed_names}")
 
 
 class Publisher(ABC):
@@ -348,7 +335,7 @@ class ItemPublisher(Publisher):
         self.post_publish_all()
 
         if errors:
-            raise PublishError(errors)
+            raise PublishError(errors, logger)
 
     def publish_one(self, item_name: str, _item: "Item") -> None:
         """


src/fabric_cicd/_items/__init__.py
```diff
@@ -1,62 +1,10 @@
...
-    "publish_orgapps",
-    "publish_reports",
-    "publish_semanticmodels",
-    "publish_sparkjobdefinitions",
-    "publish_sqldatabases",
-    "publish_userdatafunctions",
-    "publish_variablelibraries",
-    "publish_warehouses",
-    "set_unpublish_order",
+    "ItemPublisher",
```
mdrakiburrahman, line 8: The idea is to scope what we expose to these classes and avoid exposing internal functions. That way, all items maintain consistent behavior.

src/fabric_cicd/_items/_lakehouse.py
```diff
@@ -229,3 +94,148 @@ def replace_default_lakehouse_id(shortcut: dict, item_obj: Item) -> dict:
...
+        """Publish shortcuts after all lakehouses are published to protect interrelationships."""
+        if FeatureFlag.ENABLE_SHORTCUT_PUBLISH.value in constants.FEATURE_FLAG:
+            for item_obj in self.fabric_workspace_obj.repository_items.get(self.item_type, {}).values():
+                # Check if the item is published to avoid any post publish actions
+                if not item_obj.skip_publish:
+                    shortcut_publisher = ShortcutPublisher(self.fabric_workspace_obj, item_obj)
+                    shortcut_publisher.publish_all()
+
+
+class ShortcutPublisher(Publisher):
```
mdrakiburrahman, line 140: We pop lakehouses and shortcuts into a publisher class such that both can be parallelized, although there's a parent child relationship between them, they are both publishers.

src/fabric_cicd/_items/_environment.py
```diff
@@ -280,3 +245,34 @@ def _convert_environment_compute_to_camel(fabric_workspace_obj: FabricWorkspace,
...
+            return
+        _publish_environment_metadata(self.fabric_workspace_obj, item_name)
+
+    def pre_publish_all(self) -> None:
+        """Check environment publish state before publishing."""
+        _check_environment_publish_state(self.fabric_workspace_obj, True)
+
+    def post_publish_all_check(self) -> None:
+        """Check environment publish state after all environments have been published."""
+        _check_environment_publish_state(self.fabric_workspace_obj, False)
```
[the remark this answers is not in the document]
mdrakiburrahman: Yup, that was as suggested by @JacobKnightley in a previous iteration of this PR:

https://github.com/microsoft/fabric-cicd/pull/722#discussion_r2700598357

image

src/fabric_cicd/_common/_validate_input.py
```diff
@@ -156,3 +157,32 @@ def validate_token_credential(input_value: TokenCredential) -> TokenCredential:
     validate_data_type("TokenCredential", "credential", input_value)
 
     return input_value
+
+
+def validate_items_to_include(items_to_include: Optional[list[str]], operation: "OperationType") -> None:
```
shirasassoon: I like this idea of moving the check to the validate file, can we maybe do the same for other optional parameters requiring feature flags in publish/unpublish?

shirasassoon: Maybe make it a more generic function for the other parameters versus having one validate func per parameter

docs/how_to/optional_feature.md
```diff
@@ -24,6 +24,7 @@ For scenarios that aren't supported by default, fabric-cicd offers `feature-flag
 | `enable_config_deploy`                    | Set to enable config file-based deployment                              | ☑️           |
 | `enable_response_collection`              | Set to enable collection of API responses during publish operations     |              |
 | `continue_on_shortcut_failure`            | Set to allow deployment to continue even when shortcuts fail to publish |              |
+| `enable_debug_mode`                       | Set to enable debug mode for detailed logging                           |              |
```
shirasassoon: Unclear on what this is actually used for. If it is internal then we don't need to include it as part of the table, I would suggest mentioning it in the troubleshooting documentation page if relevant.

src/fabric_cicd/_common/_validate_input.py
mdrakiburrahman: Will do

docs/how_to/optional_feature.md
mdrakiburrahman: Ah this was a feature flag that I found was available in the codebase but it wasn't documented, so I thought to include it (it's the only one that was missing).

mdrakiburrahman: I just realized that this is only referenced in one of the tests and doesn't actually do anything:

image

Let me modify the test to use an existing flag, and remove this from the docs.

mdrakiburrahman: Done, removed unused FF

src/fabric_cicd/_common/_validate_input.py
mdrakiburrahman: Made a more generic `validate_experimental_param` that checks feature flags and used it in publish/unpublish

mdrakiburrahman:
@shirasassoon - Please see the changes to the logger for legibility despite parallelism:

Before:
image

Now:
image

src/fabric_cicd/_common/_file_lock.py
```diff
@@ -0,0 +1,64 @@
...
+
+import sys
+from pathlib import Path
+from types import TracebackType
+from typing import Callable, Optional, TypeVar
+
+T = TypeVar("T")
+
+
+class FileLock:
```
mdrakiburrahman, line 14: When testing with the `_http_tracer.save()` I found it wasn't thread safe when we do parallel flushing. This solves the problem.

.gitignore
```diff
@@ -164,4 +164,5 @@ cython_debug/
 
 # http traces should only be committed at the fixture root
 /http_trace.json
+/http_trace.json.lock
```
mdrakiburrahman, line 167: Since this PR introduces concurrency that contends flushing to the trace file, the lock is a persistent semaphore.

src/fabric_cicd/_common/_file_lock.py
mdrakiburrahman: Note that this filelock class is _not_ during Customer usage, this is only used for test snapshot generation.

PR #794, opened 2026-02-04
Add missing change log entries

shirasassoon:
This pull request introduces a fix that adds caching for item attribute lookups, which should improve performance and reduce redundant computations.

- Performance improvement:
  * Added caching for item attribute lookups to optimize repeated access and improve efficiency. (.changes/unreleased/fixed-20260204-150115.yaml, [.changes/unreleased/fixed-20260204-150115.yamlR1-R8](diffhunk://#diff-0088a62e42bd0588cec6dd4bfa62f8c3b6f95a3351cc481c0387f634cd8cc185R1-R8))

### DIFF
--- a/.changes/unreleased/added-20260204-152349.yaml
+++ b/.changes/unreleased/added-20260204-152349.yaml
@@ -1,0 +1,8 @@
+kind: added
+body: Add a black-box REST API testing harness
+time: 2026-02-04T15:23:49.0207511+02:00
+custom:
+    Author: mdrakiburrahman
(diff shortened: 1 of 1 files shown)


.changes/unreleased/added-20260204-152349.yaml
ayeshurun, line 1: @shirasassoon does it bring new value or fixes existing issue? if it is only for tests/dev purposes, i recommend not including it (since it is not relevant for the SDK version)

ayeshurun (approved):

PR #829, opened 2026-02-22
Fixes #811 - Disable version check on startup

### LINKED ISSUE #811: [FEATURE] Disable version check on startup
### Use Case / Problem

# What problem are you trying to solve?
We'd like to run fabric-cicd within a pipeline whose runners are heavily firewalled. The packages that we execute within the pipline are either already preloaded into the runner, or are installed on demand from an internal package registry/mirror.

# Current limitations or pain points
Within `_check_utils.py`, in `check_verison()` a call is made to pypi.org to pull the latest version number of the fabric-cicd package.

Reference: https://github.com/microsoft/fabric-cicd/blob/c26bb25754cd6d312a526b79d910b9fcccf37efd/src/fabric_cicd/_common/_check_utils.py#L58

This call is always made as it is called via the package's __init__ routine:

https://github.com/microsoft/fabric-cicd/blob/c26bb25754cd6d312a526b79d910b9fcccf37efd/src/fabric_cicd/__init__.py#L56

In a best case scenario, in a firewalled environment, this call will fail or eventually time-out. This can be a nusiance but can increase CI/CD times while it's waiting to give up ont he connection.

In a worst case scenario, watchdog-type processes such as Cimon will kill any processes attempting to make external connections that are not on an allow-list. For us, allowing pypi.org into an allow list would be out of the question as that would allow pulling in packages via channels other than our intended ones.

Give the enterprise-leaning audience for Fabric, I think it is a reasonable expectation to be able to execute the fabric-cicd package without making network calls to any unnecessary 3rd parties.

### Proposed Solution

I think a simple flag would work here, either a `--no-version-check` flag when calling the package or checking if an environment variable such as `FABRIC_CICD_NO_VERISON_CHECK` is set to true.

This can be checked in the init routine, and skip the call to `check_version()` if set. 

Either/or would work, but in my experience the environment variable path would be more desirable as we could set that on the pipeline's runner's configuration once and not have to worry about it anymore.

### Alternatives Considered

I'm planning on DIYing a fix to this internally, where fork and maintain our own version with the version check removed. This works _right now_ but is not desirable long term as we now have to maintain and keep a copy updated. It also does nothing to help the broader community.

### Impact Assessment

- [ ] This would help me personally
- [ ] This aligns with Microsoft Fabric roadmap items

### Implementation Notes

_No response_

Ricapar:
## Description

Fixes #811 

As described in issue #811, the package makes a call to pypi.org during its initalization which can create issues in firewalled or heavily isolated pipeline runtime environments.

This PR adds support for setting an environment variable, `FABRIC_CICD_VERSION_CHECK_DISABLED`, which bypasses this check.

## Example Usage and Testing

This change was tested locally via the following process:

A simple test file. Log levels are set to DEBUG in order to see urllib3's requests.

```python
#!/usr/bin/env python
import logging
logging.basicConfig(level=logging.DEBUG)
import fabric_cicd
print("fabric_cicd loaded")
```

```shell
# Run without any environment variables set
$ python test_no_version_check.py
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): pypi.org:443
DEBUG:urllib3.connectionpool:https://pypi.org:443 "GET /pypi/fabric-cicd/json HTTP/1.1" 200 12052
fabric_cicd loaded
```

The original behvior remains, and the package makes the call to pypi.org to perform the version check.

```shell
# Run with version checking disabled
$ FABRIC_CICD_VERSION_CHECK_DISABLED=1 python test_no_version_check.py
fabric_cicd loaded
```

With the environment variable set, no version check is is performed.

## Considerations made when opening this PR

* There was already an established pattern of managing the lookup of environment variables within `src/fabric_cicd/constants.py`. I added the new environment variable here following the existing naming convention.
* A constant for `VALID_ENABLE_FLAGS` was already defined and used once within the project, so I found it logical to reuse it for this check as well to maintain consistency.
* I could not find a way to add a formal unit test case for this setting that would fit within the existing testing harnesses.
* I did not find any existing documentation for the use of other environment variables, otherwise I would've added this to that. (I'd be happy to open a new issue for this later!)

### DIFF
--- a/src/fabric_cicd/constants.py
+++ b/src/fabric_cicd/constants.py
@@ -33,6 +33,8 @@ class EnvVar(str, Enum):
     """Override base delay for item name conflict retries. Defaults to 30 seconds."""
     RETRY_MAX_DURATION_SECONDS = "FABRIC_CICD_RETRY_MAX_DURATION_SECONDS"
     """Override max duration for item name conflict retries. Defaults to 300 seconds."""
+    PARALLEL_MAX_WORKERS = "FABRIC_CICD_PARALLEL_MAX_WORKERS"
+    """Override max parallel workers for concurrent item publishing. Defaults to 8."""
     VERSION_CHECK_DISABLED = "FABRIC_CICD_VERSION_CHECK_DISABLED"
     """Set to '1', 'true', or 'yes' to skip version check at startup."""
 
@@ -162,6 +164,12 @@ FABRIC_API_ROOT_URL = os.environ.get(EnvVar.FABRIC_API_ROOT_URL.value, "https://
 RETRY_AFTER_SECONDS = float(os.environ.get(EnvVar.RETRY_AFTER_SECONDS.value, 300))
 RETRY_BASE_DELAY_SECONDS = float(os.environ.get(EnvVar.RETRY_BASE_DELAY_SECONDS.value, 30))
 RETRY_MAX_DURATION_SECONDS = int(os.environ.get(EnvVar.RETRY_MAX_DURATION_SECONDS.value, 300))
+
+# Parallel Settings
+_parallel_max_workers_raw = os.environ.get(EnvVar.PARALLEL_MAX_WORKERS.value)
+PARALLEL_MAX_WORKERS: int = (
+    int(_parallel_max_workers_raw) if _parallel_max_workers_raw and _parallel_max_workers_raw.isdigit() else 8
+)
 
 # HTTP Headers
 AUTHORIZATION_HEADER = "authorization"
--- a/.changes/unreleased/20260223-115012.yaml
+++ b/.changes/unreleased/20260223-115012.yaml
@@ -1,0 +1,7 @@
+kind: fixed
+body: Add max workers soft cap with override and handle garbage response from Fabric Data Plane
+custom:
+  Author: mdrakiburrahman
+  AuthorLink: https://github.com/mdrakiburrahman
+  Issue: "827"
+  IssueLink: https://github.com/microsoft/fabric-cicd/issues/827
--- a/src/fabric_cicd/_common/_fabric_endpoint.py
+++ b/src/fabric_cicd/_common/_fabric_endpoint.py
@@ -119,6 +119,19 @@ class FabricEndpoint:
                 if logger.isEnabledFor(logging.DEBUG):
                     logger.debug(invoke_log_message)
 
+            except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as e:
+                iteration_count += 1
+                if max_duration is not None and time.time() - start_time >= max_duration:
+                    logger.debug(invoke_log_message)
+                    raise InvokeError(e, logger, invoke_log_message) from e
+                handle_retry(
+                    attempt=iteration_count,
+                    base_delay=10,
+                    max_duration=max_duration,
+                    start_time=start_time,
+                    prepend_message="Connection error encountered.",
+                )
+
             except Exception as e:
                 logger.debug(invoke_log_message)
                 raise InvokeError(e, logger, invoke_log_message) from e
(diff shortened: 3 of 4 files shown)


Ricapar:

src/fabric_cicd/constants.py
```diff
@@ -164,6 +166,9 @@ class OperationType(str, Enum):
 # HTTP Headers
 AUTHORIZATION_HEADER = "authorization"
 
+# Version Check
+VERSION_CHECK_DISABLED = bool(os.environ.get(EnvVar.VERSION_CHECK_DISABLED.value, "false") in VALID_ENABLE_FLAGS)
```
shirasassoon: Preferable to use the following:

```VERSION_CHECK_DISABLED = os.environ.get(EnvVar.VERSION_CHECK_DISABLED.value, "").lower() in VALID_ENABLE_FLAGS```

**Review by shirasassoon — COMMENTED** [review]
We need to refactor the documentation a bit so you can leave it out for now.

File: src/fabric_cicd/constants.py (hunk 1 above)
Comment (Ricapar) [reply src/fabric_cicd/constants.py]: I like this better! Thanks - the PR should be updated.

**Comment by Ricapar** [comment]
I'm going to leave it to someone who's spend a lot more time with this repository in general to better opine on the testing routine. It's a bit heavy for my own tastes given what we're testing, and it doesn't easily fit the style of the rest of the tests within this project.

If you're planning on opening a future issue on how to refactor the documentation I'd be very happy to partake in that. I agree though on that we shouldn't hijack this thread with a documentation refactor.

**Comment by JacobKnightley** [comment]
@shirasassoon there is an argument on if this version check is still beneficial.  When we were iterating frequently early on, this was helpful as even a week could mean you're behind.  Now that the library is officially supported, maybe this isn't as critical.

**Comment by shirasassoon** [comment]
> @shirasassoon there is an argument on if this version check is still beneficial. When we were iterating frequently early on, this was helpful as even a week could mean you're behind. Now that the library is officially supported, maybe this isn't as critical.

It's a good thought, thanks! I think we will keep it for now while supporting the opt-out option, I think it can still be valuable for users who are not closely following the releases even though it is now switched to a monthly cadence.

File: .changes/unreleased/added-20260221-190555.yaml [hunk 2]
```diff
@@ -0,0 +1,8 @@
+kind: added
+body: Add check for FABRIC_CICD_VERSION_CHECK_DISABLED environment variable to disable version checking on startup
```
shirasassoon: nit - rephrase the change description to:

Add FABRIC_CICD_VERSION_CHECK_DISABLED environment variable to disable version check on startup

Ricapar:
> @shirasassoon there is an argument on if this version check is still beneficial. When we were iterating frequently early on, this was helpful as even a week could mean you're behind. Now that the library is officially supported, maybe this isn't as critical.

Funny enough, I had prepared a very different post earlier on but had opted not to go for it, but since you mentioned it..

I agree that longer-term you should consider removing the verison check. I don't think the behavior is appropriate for a library that is then included in other's Python scripts.

I can understand the behavior in a full CLI tool (take the `fab` Fabric CLI for instance) where it'll perform a version check on startup (though I would also make sure that in that scenario you still give a way to override the check!). 

I don't think an external network call on load is appropriate for what essentially bundles and ships as a standalone Python library that you then `include fabric_cicd` into your own script.

shirasassoon:
> > @shirasassoon there is an argument on if this version check is still beneficial. When we were iterating frequently early on, this was helpful as even a week could mean you're behind. Now that the library is officially supported, maybe this isn't as critical.
> 
> Funny enough, I had prepared a very different post earlier on but had opted not to go for it, but since you mentioned it..
> 
> I agree that longer-term you should consider removing the verison check. I don't think the behavior is appropriate for a library that is then included in other's Python scripts.
> 
> I can understand the behavior in a full CLI tool (take the `fab` Fabric CLI for instance) where it'll perform a version check on startup (though I would also make sure that in that scenario you still give a way to override the check!).
> 
> I don't think an external network call on load is appropriate for what essentially bundles and ships as a standalone Python library that you then `include fabric_cicd` into your own script.

Thanks for the feedback!

PR #853, opened 2026-03-05
Closes #695 Expose API responses in deploy_with_config for both success and failure scenarios

### LINKED ISSUE #695: Expose API responses in deploy_with_config for both success and failure scenarios
### What is the feature?

**The Gap:**
When using `deploy_with_config`, the `FabricWorkspace` object is instantiated internally within the function scope.
1.  **On Success:** The function returns `None`, swallowing the `responses` dictionary (when `enable_response_collection` is active).
2.  **On Failure:** The function raises an exception immediately. Because the `FabricWorkspace` object is local to the function scope, it is destroyed when the stack unwinds. The caller has **no way** to access `workspace.responses` to determine which items succeeded before the crash.

This makes `deploy_with_config` unsuitable for production pipelines where reporting partial success (e.g., "Notebooks deployed, but Pipelines failed") is required.

**The Ideal Solution:**
Modify `deploy_with_config` in `src/fabric_cicd/publish.py` to wrap the operations in a `try...except` block.
1.  **Happy Path:** Return `workspace.responses`.
2.  **Failure Path:** Catch the exception, attach the `workspace.responses` dictionary to the Exception instance (e.g., as a `.partial_results` attribute), and re-raise the exception.

**Proposed Implementation Logic:**
```python
# src/fabric_cicd/publish.py

def deploy_with_config(...):
    # ... setup workspace ...
    
    try:
        # Execute deployment operations
        if not publish_settings.get("skip", False):
            publish_all_items(workspace, ...)
            
        if not unpublish_settings.get("skip", False):
            unpublish_all_orphan_items(workspace, ...)
            
    except Exception as e:
        # Attach partial responses to the exception object before re-raising
        # This preserves the data that would otherwise be lost when 'workspace' goes out of scope
        if hasattr(workspace, 'responses') and workspace.responses:
            # Dynamically attach attribute to exception instance
            e.partial_results = workspace.responses
        raise e
        
    # Happy path return
    return getattr(workspace, 'responses', None)
```

### Additional context

**Example Usage if implemented:**

```python
from fabric_cicd import deploy_with_config, append_feature_flag

append_feature_flag("enable_response_collection")
# ... enable other flags ...

try:
    results = deploy_with_config(config_file_path="config.yml", environment="PROD")
    print("Deployment Success!")
    # Process results (e.g. log new Item IDs)
    if results:
        print(results)

except Exception as e:
    print("Deployment Failed!")
    
    # Feature allows accessing partial data attached to the exception
    if hasattr(e, 'partial_results'):
        print("Items that succeeded before crash:")
        for item_type, items in e.partial_results.items():
            print(f"- {item_type}: {list(items.keys())}")
            
    raise e # Fail the pipeline
```

shirasassoon:
This pull request enhances the transparency and robustness of the `deploy_with_config` function by introducing structured reporting of deployment outcomes, including both success and failure cases. It adds support for collecting and exposing API response data when enabled via feature flags, and ensures that failures are surfaced with detailed context. Comprehensive tests have been added to verify these behaviors.

Improvements to deployment result reporting:

* The `DeploymentResult` dataclass now includes a `responses` field (optional dictionary) for API response data, and is returned on both success and failure scenarios. On failure, an instance is attached to the exception as `e.deployment_result` with `status` set to `FAILED`. (`src/fabric_cicd/_common/_deployment_result.py`, [src/fabric_cicd/_common/_deployment_result.pyL6-R36](diffhunk://#diff-6b1f126a473045dc8ef68337500a5eda56d5a1e190727d06f1d3b13baecfef79L6-R36))
* The `deploy_with_config` function is updated to return `DeploymentResult` with `responses` populated when the `enable_response_collection` feature flag is enabled, and attaches partial responses to exceptions on failure. (`src/fabric_cicd/publish.py`, [[1]](diffhunk://#diff-a8555a82747092e8d1f2ee36f31c947b4e5a264ed6c6d8de48d8daf850372704L347-R362) [[2]](diffhunk://#diff-a8555a82747092e8d1f2ee36f31c947b4e5a264ed6c6d8de48d8daf850372704R402-R425) [[3]](diffhunk://#diff-a8555a82747092e8d1f2ee36f31c947b4e5a264ed6c6d8de48d8daf850372704R439-R441) [[4]](diffhunk://#diff-a8555a82747092e8d1f2ee36f31c947b4e5a264ed6c6d8de48d8daf850372704R474-R489)

Feature flag integration:

* Response collection is now controlled by the `enable_response_collection` feature flag in the config file, allowing users to opt-in for richer deployment diagnostics. (`src/fabric_cicd/publish.py`, [src/fabric_cicd/publish.pyR439-R441](diffhunk://#diff-a8555a82747092e8d1f2ee36f31c947b4e5a264ed6c6d8de48d8daf850372704R439-R441))

Testing and validation:

* New and updated tests validate that response collection works as expected, that exceptions include deployment results and partial responses, and that feature flags are properly scoped and restored. (`tests/test_deploy_with_config.py`, [[1]](diffhunk://#diff-591b837760640c6d5c7b21d70575e6a262d56881f9ff3f993dbf10064c61d468R534-R610) [[2]](diffhunk://#diff-591b837760640c6d5c7b21d70575e6a262d56881f9ff3f993dbf10064c61d468L1162-R1280) [[3]](diffhunk://#diff-591b837760640c6d5c7b21d70575e6a262d56881f9ff3f993dbf10064c61d468R1508-R1694) [[4]](diffhunk://#diff-591b837760640c6d5c7b21d70575e6a262d56881f9ff3f993dbf10064c61d468L1260-L1302)

Documentation and changelog:

* Added detailed docstrings and usage examples to `deploy_with_config`, and updated the changelog to reflect improved transparency in deployment outcomes. (`src/fabric_cicd/publish.py`, [[1]](diffhunk://#diff-a8555a82747092e8d1f2ee36f31c947b4e5a264ed6c6d8de48d8daf850372704L347-R362); `.changes/unreleased/added-20260312-113803.yaml`, [[2]](diffhunk://#diff-ac6eb083fc9b526e342b4d17288bc963de7dafc847bd2b2a1e7af62a9d2ab21eR1-R8)

### DIFF
--- a/src/fabric_cicd/publish.py
+++ b/src/fabric_cicd/publish.py
@@ -351,8 +351,8 @@ def deploy_with_config(
             otherwise None.
 
     Raises:
-        InputError: If configuration file is invalid or environment not found.
-        FileNotFoundError: If configuration file doesn't exist.
+        InputError: If configuration is invalid, environment not found, or input validation fails.
+        ConfigValidationError: If configuration file is missing or fails structural validation.
 
     Note:
         On failure, the raised exception will have a ``deployment_result`` attribute


src/fabric_cicd/publish.py
```diff
@@ -389,57 +399,92 @@ def deploy_with_config(
...
+            responses=partial_results,
         )
+        raise
+
+    logger.info("Config-based deployment completed successfully")
+    return DeploymentResult(
+        status=DeploymentStatus.COMPLETED,
+        message="Deployment completed successfully",
+        responses=(workspace.responses if responses_enabled and workspace.responses else None),
+    )
```
ayeshurun, line 488: What is the reason for this feature flag? Why not always returning the responses? It's not a breaking change

src/fabric_cicd/publish.py
```diff
@@ -389,57 +399,92 @@ def deploy_with_config(
...
+                    item_name_exclude_regex=unpublish_settings.get("exclude_regex", "^$"),
+                    items_to_include=unpublish_settings.get("items_to_include"),
+                )
+            else:
+                logger.info(f"Skipping unpublish operation for environment '{environment}'")
+
+    except Exception as e:
+        partial_results = None
+        if responses_enabled and workspace is not None:
+            partial_results = getattr(workspace, "responses", None) or None
```
aviatco: I think this - getattr(workspace, "responses", None), return None in case no workspace so the "or None" part is redundant

src/fabric_cicd/publish.py
```diff
@@ -389,57 +399,92 @@ def deploy_with_config(
...
+        e.deployment_result = DeploymentResult(
+            status=DeploymentStatus.FAILED,
+            message=str(e),
+            responses=partial_results,
```
aviatco: Do we need to update the CLI result?

shirasassoon: No CLI is returning the str(e) anyways so CLI doesn't need to change

src/fabric_cicd/publish.py
shirasassoon: fixed

src/fabric_cicd/publish.py
shirasassoon: It exposes additional info (API responses) which may or may not be desired by the user to see so wanted to make it an available feature, but opt-in. This was when we initially introduced it - open to discuss if we want to change this to the default behavior.

PR #897, opened 2026-03-22
Add custom agent instructions for new item type onboarding

shirasassoon:
create custom agent instructions for onboarding new item types in fabric-cicd.

### DIFF
--- a/.github/agents/new-item-type.agent.md
+++ b/.github/agents/new-item-type.agent.md
@@ -14,17 +14,7 @@ tools:
 
 # New Item Type Onboarding Agent
 
-You are an expert at onboarding new Microsoft Fabric item types into the `fabric-cicd` Python library. You guide contributors through every integration point, generate the correct code, and validate completeness.
-
-> **Important:** If you are unsure about any detail — such as whether the item type supports definitions, has unique deployment requirements, depends on other item types, or requires parameterization — **always ask the requestor for clarification before proceeding**. Do not guess or assume. It is better to pause and confirm than to generate incorrect code.
-
-## When to Use This Agent
-
-Use this agent when you need to:
-
-- Add support for a brand-new Fabric item type (e.g., `Ontology`, `Map`)
-- Understand what files need to change to register a new item type
-- Generate the boilerplate code for a new item type end-to-end
+> **Important:** If you are unsure about any detail — such as whether the item type supports definitions, has unique deployment requirements, depends on other item types, or requires parameterization — **always ask the requestor for clarification before proceeding**. List the specific unknowns and ask the requestor to confirm each one before continuing.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


shirasassoon:
> **1. "You are an expert" preamble (Line 17)**
> 
> The YAML frontmatter (name, description) already establishes the agent's identity and scope for the LLM. Repeating it in prose creates two sources of truth — if one gets updated and the other doesn't, the LLM receives conflicting signals. This line can be removed, or replaced with a single sentence that adds context the frontmatter doesn't cover (e.g., scope boundaries or a constraint).
> 
> **2. "When to Use This Agent" section (Lines 21–28)**
> 
> IF the user selects this agent explicitly (via @new-item-type or agent picker) — the LLM doesn't decide whether to invoke itself. The description and argument-hint in frontmatter are the discoverability mechanisms for the user. This section consumes tokens without influencing the agent's behavior. Removing it keeps the instruction set focused on what to do, not when to be used.
> 
> **3. Safety Rules at the bottom (Lines 297–307)**
> 
> LLMs process instructions sequentially and tend to weight earlier content more strongly, especially in long documents. Placing safety constraints after 300 lines of procedural steps means the agent may start generating code before encountering the rules that should govern how it generates code. Moving safety rules before the integration checklist ensures they're loaded into context before the agent begins executing steps.
> 
> **4. No explicit phasing between gathering info and writing code**
> 
> The agent has good prerequisite and eligibility gate sections, but there's no explicit instruction telling the LLM to stop and complete gathering before starting code generation. Without that gate, the LLM may short-circuit — it sees the first few prereqs answered and jumps ahead to Step 1. Adding a brief phased workflow (e.g., "Complete all prerequisites and eligibility gates before proceeding to the integration checklist") gives the LLM a clear decision boundary.
> 
> **5. Validation checklist mixes actionable and non-actionable items (Lines 261–275)**
> 
> The final three items (uv run python -c ..., uv run pytest -v, uv run ruff format/check) are runnable — the agent can execute those and confirm pass/fail. The structural checks above them (lines 261–271) are assertions the LLM can only state, not verify. For example, "ItemType.NEW_TYPE exists in the enum" could become grep -n 'NEW_TYPE' src/fabric_cicd/constants.py — giving the agent something to run rather than assert. Converting the structural items to verification commands makes the entire checklist executable end-to-end.
> 
> **6. Patterns table could be more prescriptive about step scoping (Lines 243–253)**
> 
> The "Patterns and Reference Examples" table maps patterns to steps and example files — this is a strong reference. One gap: it tells the LLM what additional steps each pattern adds, but doesn't tell it which steps to skip. For a "Simple" item, the agent may still walk through all 6 steps because nothing explicitly says "Steps 1c, 1d, 1e, 1f, 1g do not apply." Adding a brief "skip steps" column or a classification table that maps complexity level to the exact step set would reduce unnecessary work and hallucinated output.
> 
> **7. Emoji in prerequisite table (Lines 37–41)**
> 
> Minor: ✅ in the "Required" column renders inconsistently across terminal emulators and diff viewers that contributors may use. Plain Yes is more portable for a developer-facing document.
> 
> **8. Important callout has redundant softening language (Line 19)**
> 
> The Important callout ends with: "Do not guess or assume. It is better to pause and confirm than to generate incorrect code." The actionable instruction is already in the sentence before it: "always ask the requestor for clarification before proceeding." The trailing sentences restate that same instruction in vaguer, softer terms — "pause and confirm" doesn't tell the LLM what to confirm or how to pause, and "do not guess or assume" is a negative instruction without a concrete alternative (the concrete alternative was already stated). LLMs respond more reliably to specific affirmative instructions than to "don't do X" phrasing. The redundancy also costs tokens without adding signal. Consider trimming to just the bolded instruction, or replacing the trailing text with something specific and additive — for example, listing what the agent should surface to the user when it encounters ambiguity (e.g., "List the specific unknowns and ask the requestor to confirm each one before continuing").

@jeremydhoover-blip Thanks for the feedback. Addressed all the points, please feel free to review or approve the PR.

PR #907, opened 2026-03-26
Fixes #866 Ignore paramater file when not explicitly defined in config file

### LINKED ISSUE #866: [BUG] Ignore parameter file when not explicitly defined in config file
### Library Version

Latest

### Python Version

All

### Operating System

Windows

### Authentication Method

Other

### What is the problem?

When the parameter field in the config file is omitted, if a parameter.yml file is present in the repository directory it will be picked up by fabric-cicd and used during the deployment.

### Steps to reproduce

1. Use the `config.yml` file under sample and update workspace and other fields appropriately
2. Remove the parameter field entirely 
3. Execute the `debug_local config.py` file 
4. Notice parameter file validation step and parameterization is applied during deployment (when parameter.yml exists under sample workspace folder)

### Expected behavior

When the parameter field is omitted in the config file, parameterization should never be applied. The user must explicitly define the parameter field for parameterization to be used during the deployment. 

### Actual behavior

If there is a parameter.yml file in the repository directory and the parameter field is removed from config file, the deployment uses the parameter file and applies parameterization. 

Pass in a kwarg from `deploy_with_config` to `FabricWorkspace` that skips parameterization when no parameter file path is defined in the config file. 

_No response_

aviatco:
This pull request refines the logic for parameter file handling and parameterization in the deployment workflow. The main change is to require explicit configuration for parameterization: a parameter file will only be loaded and applied if the user explicitly specifies it in the config. Otherwise, even if a `parameter.yml` file exists in the repository, it will be ignored. This prevents accidental or implicit parameterization and clarifies the deployment behavior. The PR also adds and updates tests to verify this stricter behavior.

**Parameterization logic changes:**

* The `extract_workspace_settings` function now only sets the `parameter_file_path` in the settings if the `parameter` field is explicitly defined in the config. If omitted, parameterization is disabled and the absence of `parameter_file_path` signals this.
* In `fabric_workspace.py`, the parameter file is only validated and loaded if an explicit `parameter_file_path` is provided. If not, parameterization is skipped and a log message is emitted. [[1]](diffhunk://#diff-3bd0f70ed06c7fc7a0f77378aecc6fb108eeb5263161db3b22270f3de4df99c0L245-R254) [[2]](diffhunk://#diff-3bd0f70ed06c7fc7a0f77378aecc6fb108eeb5263161db3b22270f3de4df99c0R268-R269)
* The `_set_parameter_file_path` method in the `Parameter` class is simplified to remove default path auto-discovery—parameterization is only enabled when explicitly configured.

**Testing and validation:**

* New and updated tests confirm that parameter files are not auto-discovered or applied when not explicitly configured, and that parameterization only occurs when the path is provided. This includes tests for both presence and absence of the `parameter` field and the `parameter_file_path`. [[1]](diffhunk://#diff-591b837760640c6d5c7b21d70575e6a262d56881f9ff3f993dbf10064c61d468R270-R297) [[2]](diffhunk://#diff-38a69298c91571d1f1a72285d551b8e0ceaebe1ba66bd7bfc7a537620b4bfa3aR928-R972)

**Documentation and minor clarifications:**

* Docstrings and comments are updated to clarify that parameterization is opt-in and not applied by default.

These changes make parameterization behavior explicit and predictable, reducing the risk of accidental configuration and improving test coverage.## Description

Briefly describe what this PR does and why.

## Linked Issue (REQUIRED)
#866

### DIFF
--- a/.changes/unreleased/fixed-20260325-124736.yaml
+++ b/.changes/unreleased/fixed-20260325-124736.yaml
@@ -1,0 +1,8 @@
+kind: fixed
+body: Fix Notebook deployment failure caused by non-deterministic ordering of definition files in API payload
+time: 2026-03-25T12:47:36.1355405+02:00
+custom:
+    Author: shirasassoon
+    AuthorLink: https://github.com/shirasassoon
+    Issue: "869"
+    IssueLink: https://github.com/microsoft/fabric-cicd/issues/869
--- a/.github/agents/new-item-type.agent.md
+++ b/.github/agents/new-item-type.agent.md
@@ -1,0 +1,319 @@
+---
+name: New Item Type
+description: Guide and assist with onboarding a new Microsoft Fabric item type into fabric-cicd
+argument-hint: Tell me which Fabric item type you want to add (e.g., "Add support for Ontology")
+tools:
+    - runInTerminal
+    - terminalLastCommand
+    - search
+    - fetch
+    - readFile
+    - editFiles
+    - createFile
+---
+
+# New Item Type Onboarding Agent
+
+> **Important:** If you are unsure about any detail — such as whether the item type supports definitions, has unique deployment requirements, depends on other item types, or requires parameterization — **always ask the requestor for clarification before proceeding**. List the specific unknowns and ask the requestor to confirm each one before continuing.
+
+## Prerequisites
+
+Before starting, gather the following information about the new item type:
+
+### Core Information (always gather before starting)
+
+| Information                                                | Example           | Required |
+| ---------------------------------------------------------- | ----------------- | -------- |
+| **Display name** (PascalCase, as used by Fabric API)       | `CopyJob`         | Yes      |
+| **Supported in source control / Git integration**          | Yes / No          | Yes      |
+| **Fabric API supports deployment** (create/update via API) | Yes / No          | Yes      |
+| **Deployment type** (full definition or shell-only)        | Full / Shell-only | Yes      |
+| **Supports service principal (SPN) authentication**        | Yes / No          | Yes      |
+
+### Additional Details (gather when relevant to the item type)
+
+| Information                                     | Example                                             | Detail                                                                             |
+| ----------------------------------------------- | --------------------------------------------------- | ---------------------------------------------------------------------------------- |
+| **Alternate Definition format**                 | `ipynb`, `SparkJobDefinitionV2`                     | If the item uses a non-standard API format                                         |
+| **Custom deployment logic**                     | Creation payload, post-publish binding, async check | If the item needs special handling beyond standard publish                         |
+| **Custom file content transformation**          | Rewrite cross-item refs, inject deployed values     | If file contents need item-type-specific changes the generic pipeline can't handle |
+| **Dependencies on other item types**            | Eventhouse → KQLDatabase                            | If the item depends on another item type existing first                            |
+| **Destructive unpublish** (data loss on delete) | Lakehouse, Eventhouse                               | If deleting the item destroys user data                                            |
+| **Exclude paths during publish**                | `.pbi/`, `.children/`                               | If certain files within the item folder should be skipped                          |
+| **Intra-type dependencies**                     | Pipeline invokes another pipeline                   | If items of this type can reference each other                                     |
+| **Specialized parameterization**                | `spark_pool`, `semantic_model_binding`              | If the item needs a dedicated `parameter.yml` key beyond generic find/replace      |
+
+### Eligibility Gates
+
+Before proceeding, confirm all of the following. If any gate fails, **stop** — the item type cannot be onboarded.
+
+1. The item type must be supported in source control / Git integration. See [supported item types for Git integration](https://learn.microsoft.com/en-us/rest/api/fabric/articles/item-management/definitions/item-definition-overview).
+2. The Fabric API must support deployment for the item type — either full definition deployment or shell-only creation (like Lakehouse/Warehouse). Search the [Fabric REST API docs](https://learn.microsoft.com/en-us/rest/api/fabric/) to confirm.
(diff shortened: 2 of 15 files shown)


src/fabric_cicd/fabric_workspace.py
```diff
@@ -242,25 +242,31 @@ def _get_item_attribute(
...
-            parameter_file_name=constants.PARAMETER_FILE_NAME,
-            parameter_file_path=self.parameter_file_path,
-        )
-        is_valid = parameter_obj._validate_parameter_file()
-        if is_valid:
-            self.environment_parameter = parameter_obj.environment_parameter
+
+        # Only validate and load the parameter file when an explicit path is provided.
+        # When parameter_file_path is None, parameterization is disabled.
+        if self.parameter_file_path is not None:
```
shirasassoon: I'm not sure this logic is correct, while it considers the config-based deployment it might break the legacy deployment approach where users don't usually provide an explicit parameter file path

tests/test_deploy_with_config.py
```diff
@@ -267,6 +267,38 @@ def test_extract_parameter_file_path_missing(self):
...
+                "repository_directory": "test/path",
+                "parameter": "parameter.yml",
+            }
+        }
+
+        settings = extract_workspace_settings(config, "dev")
+        # parameter_file_path present → skip_parameterization must be False
+        skip_parameterization = settings.get("parameter_file_path") is None
+        assert skip_parameterization is False
+
```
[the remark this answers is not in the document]
aviatco: I think we should keep this since all the tests in this file are doing the same logic
@shirasassoon  what do you think?

src/fabric_cicd/fabric_workspace.py
```diff
@@ -158,8 +158,19 @@ def __init__(
...
+        skip_parameterization = kwargs.get("skip_parameterization", False)
+        if not skip_parameterization:
+            self._refresh_parameter_file()
+        else:
+            self.environment_parameter = {}
+            logger.info(
+                "Parameterization skipped: no parameter file configured/provided "
+                "(environment=%s, parameter_file_path=%s).",
+                self.environment,
+                self.parameter_file_path,
```
shirasassoon: What would `self.parameter_file_path` print in this case?

tests/test_deploy_with_config.py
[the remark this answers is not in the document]
shirasassoon: @aviatco I would recommend following this suggestion:  consider rewriting them to assert the actual behavior (e.g., that `deploy_with_config` passes the skip flag / that parameterization is skipped) rather than recomputing the flag in the test.

src/fabric_cicd/fabric_workspace.py
aviatco: nothing.. will remove it

tests/test_deploy_with_config.py
```diff
@@ -267,7 +267,6 @@ def test_extract_parameter_file_path_missing(self):
         settings = extract_workspace_settings(config, "dev")
         assert "parameter_file_path" not in settings
 
-
 class TestPublishSettingsExtraction:
```
shirasassoon, line 270: Is there any integration test that verifies that `deploy_with_config()` correctly computes `skip_parameterization=True` when the config omits parameter and passes it through?

shirasassoon:
Add a changelog entry - bug fix

tests/test_deploy_with_config.py
shirasassoon:
```
def test_deploy_with_config_skips_parameterization_when_parameter_absent(self, tmp_path):
    """Integration: deploy_with_config must not auto-discover parameter.yml
    when the 'parameter' field is absent from the config file."""
    # Set up repo directory WITH a parameter.yml that would be auto-discovered
    repo_dir = tmp_path / "workspace"
    repo_dir.mkdir()
    (repo_dir / "parameter.yml").write_text(
        "find_replace:\n"
        "  - find_value: 'old'\n"
        "    replace_value:\n"
        "      dev: 'new'\n"
    )

    config_data = {
        "core": {
            "workspace_id": {"dev": "11111111-1111-1111-1111-111111111111"},
            "repository_directory": str(repo_dir),
            # 'parameter' intentionally omitted
        }
    }
    config_file = tmp_path / "config.yml"
    config_file.write_text(yaml.dump(config_data))

    with patch("fabric_cicd.publish.FabricWorkspace") as MockWS:
        mock_ws = MagicMock()
        mock_ws.environment_parameter = {}
        MockWS.return_value = mock_ws

        # Call deploy_with_config and verify skip_parameterization=True was passed
        # (mock publish/unpublish to avoid real API calls)
        with patch("fabric_cicd.publish.publish_all_items"), \
             patch("fabric_cicd.publish.unpublish_all_orphan_items"):
            deploy_with_config(str(config_file), "dev", token_credential=MagicMock())

        # Assert FabricWorkspace was constructed with skip_parameterization=True
        call_kwargs = MockWS.call_args[1]
        assert call_kwargs.get("skip_parameterization") is True


def test_deploy_with_config_loads_parameter_when_field_present(self, tmp_path):
    """Integration: deploy_with_config must pass skip_parameterization=False
    when the 'parameter' field IS present in config."""
    repo_dir = tmp_path / "workspace"
    repo_dir.mkdir()

    config_data = {
        "core": {
            "workspace_id": {"dev": "11111111-1111-1111-1111-111111111111"},
            "repository_directory": str(repo_dir),
            "parameter": "my-params.yml",
        }
    }
    config_file = tmp_path / "config.yml"
    config_file.write_text(yaml.dump(config_data))

    with patch("fabric_cicd.publish.FabricWorkspace") as MockWS:
        mock_ws = MagicMock()
        MockWS.return_value = mock_ws

        with patch("fabric_cicd.publish.publish_all_items"), \
             patch("fabric_cicd.publish.unpublish_all_orphan_items"):
            deploy_with_config(str(config_file), "dev", token_credential=MagicMock())

        call_kwargs = MockWS.call_args[1]
        assert call_kwargs.get("skip_parameterization") is False
```

aviatco: Added under TestDeployWithConfig

aviatco:

PR #943, opened 2026-04-20
warning still appears for users, but it won’t corrupt stdout output

### LINKED ISSUE #944: [BUG] 'new version available' warning corrupts stdout output
### Library Version

0.3.1

### Python Version

3.14

### Operating System

Windows

### Authentication Method

Azure CLI (az login)

### What is the problem?

When importing/using `fabric_cicd`, `check_version()` runs on import and writes the update notice to stdout via `print(msg)` in `src/fabric_cicd/_common/_check_utils.py`. This corrupts machine-readable stdout output in automation/scripts. 

Example: structured output like `WORKSPACE_GUID: <value>` gets mixed with the multi-line version notice, breaking parsers.

### Steps to reproduce

1. Install an older `fabric-cicd` version while a newer version exists.
2. Run a script/command that imports `fabric_cicd` and prints structured stdout values.
3. Observe `[notice] A new release...` text injected into stdout.

### Expected behavior

Version update notice should not be written to stdout. It should go to logging/stderr so stdout remains parse-safe.

### Actual behavior

Replace `print(msg)` with `logger.warning(msg)` (or stderr output) in `check_version()`.

Replace `print(msg)` with `logger.warning(msg)` (or stderr output) in `check_version()`.

Image

jjmerelo:
Fixes #944

### DIFF
--- a/src/fabric_cicd/_common/_check_utils.py
+++ b/src/fabric_cicd/_common/_check_utils.py
@@ -86,7 +86,7 @@ def check_version() -> None:
                 f"{Fore.CYAN}https://microsoft.github.io/fabric-cicd/latest/changelog/{Style.RESET_ALL}"
             )
 
-            print(msg)
+            logger.warning(msg)
     except Exception as e:
         # Silently handle errors, but log them if debug is needed
         logger.debug(f"Error checking version: {e}")
--- a/.changes/unreleased/fix-version-notice-stdout.yaml
+++ b/.changes/unreleased/fix-version-notice-stdout.yaml
@@ -1,0 +1,6 @@
+kind: fixed
+body: Keep version update notices out of stdout so machine-readable command output is not corrupted
+time: 2026-04-27T12:00:00.0000000-04:00
+custom:
+  Issue: 944
+  Author: jjmerelo
--- a/tests/test__check_utils.py
+++ b/tests/test__check_utils.py
@@ -2,9 +2,12 @@
 # Licensed under the MIT License.
 
 import json
+import logging
+from unittest.mock import MagicMock
 
 import pytest
 
+import fabric_cicd._common._check_utils as check_utils
 from fabric_cicd._common._check_utils import check_file_type, check_valid_json_content, check_valid_yaml_content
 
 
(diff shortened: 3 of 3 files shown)


shirasassoon (commented):
@jjmerelo Looks like a simple fix! Let's add one focused stdout regression unit test in case any change to this is made in the future.

shirasassoon (commented):
Please add a changie entry - for instructions see [here](https://github.com/microsoft/fabric-cicd/blob/main/CONTRIBUTING.md#documenting-changes-with-changie)

jjmerelo:

jjmerelo:
@shirasassoon added the regression testing and the changie entry. please let me know if there's anything else. thank you!

src/fabric_cicd/_common/_check_utils.py
```diff
@@ -86,7 +86,7 @@ def check_version() -> None:
                 f"{Fore.CYAN}https://microsoft.github.io/fabric-cicd/latest/changelog/{Style.RESET_ALL}"
             )
 
-            print(msg)
+            logger.warning(msg)
```
shirasassoon, line 89: @jjmerelo, one small follow-up question before we merge: since the version notice message uses colorama codes (e.g. Fore.BLUE, Style.RESET_ALL), could you confirm whether you see any raw ANSI escape sequences (like \x1b[34m) appearing in your log output with this change? This isn't a blocker — just want to make sure log sinks like file handlers or CI platforms aren't showing garbled output. If everything looks clean on your end, we're good to go!

src/fabric_cicd/_common/_check_utils.py
shirasassoon, line 68: Now that we are using the logger it includes new line so we can remove the `\n` at the end of 68 to avoid double new line

shirasassoon, line 74: Same comment as above, in this case just leave one `\n`

shirasassoon, line 82: Same comment - `msg += "\n"` can be removed entirely

shirasassoon:
@jjmerelo Please see the latest review comments. Thanks.

shirasassoon:
@jjmerelo thank you again for taking the time to submit this fix. Ultimately, this PR may not be needed since we're planning to remove the version check entirely to prevent unnecessary noise and slowdowns during import.

PR #956, opened 2026-04-25
Fixes #754 - Add FQDN workspace ID transformation for private-link Fabric

### LINKED ISSUE #754: Support for using the workspace Fully Qualified Domain Name (FQDN) endpoint
### What is the feature?

As documented [here](https://learn.microsoft.com/en-us/fabric/security/security-workspace-level-private-links-overview#connecting-to-workspaces) by Microsoft, if a workspace has the setting 'Allow connections only from workspace level private links' enabled, the default fabric workspace api endpoint is no longer accessible.
e.g. any api path starting with https://api.fabric.microsoft.com/v1/workspaces/{workspace_id} is no longer accessible since that resolves to the public endpoint. 

To connect to a private workspace the fully qualified domain has to be used. e.g. https://{workspaceid}.z{xy}.w.api.fabric.microsoft.com/v1/workspaces/{workspace_id},
where {workspaced_id} is the workspace id without dashes, and {xy} is the first two characters of the workspace id.

I have enabled workspace level private link recently, and am no longer able to deploy my workspaces with fabric-cicd because of this limitation. 

I think the ideal solution would be to always use the FQDN when connecting to a specific workspace, since it resolves to the public endpoint anyway if no private link is set up. Otherwise a toggle or override option for this functionality could be made.

### Additional context

Overriding the root api url like below resolves the issue for my private link enabled workspace. This is not a clean solution but it is a workaround that shows that the fabric-cicd package works fine with the FQDN (at least in my setup)

``` python
from fabric_cicd import (
    FabricWorkspace,
    append_feature_flag,
    constants,
    publish_all_items,
    unpublish_all_orphan_items,
)

# Get the Fabric workspace ID from command line arguments
workspace_id = sys.argv[1]

# Override the API root URL
constants.DEFAULT_API_ROOT_URL = f"https://{workspace_id.replace('-', '')}.z{workspace_id[0:2]}.w.api.fabric.microsoft.com"
```

SumeshKashyap:
## Description

Description

This PR introduces support for FQDN-based workspace ID transformation for private-link-enabled Microsoft Fabric workspaces. The implementation centralizes the logic in constants.py to ensure consistent and reusable configuration of API root URLs.

Two new functions are added:

_get_fabric_fqdn_url: A helper function that converts a workspace ID into the required FQDN format by removing dashes and deriving the appropriate regional subdomain.
setup_fabric_fqdn: A public function that updates global API root URL constants (FABRIC_API_ROOT_URL and DEFAULT_API_ROOT_URL) based on the transformed FQDN.

Additionally, the new function is exposed via __init__.py for public API access.

This change enables seamless integration with private-link-enabled workspaces and ensures that API calls are routed correctly using the appropriate FQDN format.

Closes #754 

# Usage of the new function:
from fabric_cicd import FabricWorkspace, publish_all_items, unpublish_all_orphan_items, setup_fabric_fqdn

##### call function
setup_fabric_fqdn(workspace_id)

##### Initialize the FabricWorkspace object with the required parameters
target_workspace = FabricWorkspace(
    workspace_id=workspace_id,
    environment=environment,
    repository_directory=repository_directory,
    item_type_in_scope=item_type_in_scope,
    token_credential=token_credential,
)


SumeshKashyap:

shirasassoon:
@SumeshKashyap Thanks for the contribution! The linked issue was closed and concluded that overriding the API URL constant is a sufficient workaround for deploying to Private Linked Workspaces. I would like to understand why it wouldn't be sufficient for your case and why the proposed solution is better. Thanks

SumeshKashyap:
@shirasassoon 

Thank you for the feedback!

I agree that overriding the API URL constant works as a workaround, and it does unblock deployments to private-link-enabled workspaces. However, the motivation behind this proposal is to make that workaround more robust, reusable, and less error-prone in real-world usage. 

In our case when handling multiple workspace deployment pipelines, manually overriding the API URL constant introduced a few challenges:

* **Repetition & inconsistency**: The override logic had to be implemented in multiple places across scripts and pipelines, increasing the risk of inconsistencies.
* **Error-prone transformation**: Constructing the FQDN requires specific steps (removing dashes, extracting characters, formatting the URL). Repeating this logic manually increases the likelihood of subtle bugs.
* **Lack of discoverability**: Users of the library may not be aware that they need to override the constant or how to correctly derive the FQDN.
* **Maintainability concerns**: If the FQDN format evolves in the future, updating scattered overrides becomes difficult compared to a centralized implementation.

The proposed solution centralizes this logic within `constants.py` and exposes a simple public function (`setup_fabric_fqdn`) that:

* Encapsulates the transformation logic in one place
* Ensures consistent and correct FQDN generation
* Provides a clean and discoverable API for users
* Reduces duplication across projects and pipelines

From a design perspective, this moves the approach from a **manual workaround** to a **first-class, supported pattern** within the library, while still preserving backward compatibility.

Happy to adjust the approach if you’d prefer a lighter-touch solution (for example, keeping it internal-only or documenting the pattern more clearly). Would love to hear your thoughts.

shirasassoon:
> @shirasassoon
> 
> Thank you for the feedback!
> 
> I agree that overriding the API URL constant works as a workaround, and it does unblock deployments to private-link-enabled workspaces. However, the motivation behind this proposal is to make that workaround more robust, reusable, and less error-prone in real-world usage.
> 
> In our case when handling multiple workspace deployment pipelines, manually overriding the API URL constant introduced a few challenges:
> 
> * **Repetition & inconsistency**: The override logic had to be implemented in multiple places across scripts and pipelines, increasing the risk of inconsistencies.
> * **Error-prone transformation**: Constructing the FQDN requires specific steps (removing dashes, extracting characters, formatting the URL). Repeating this logic manually increases the likelihood of subtle bugs.
> * **Lack of discoverability**: Users of the library may not be aware that they need to override the constant or how to correctly derive the FQDN.
> * **Maintainability concerns**: If the FQDN format evolves in the future, updating scattered overrides becomes difficult compared to a centralized implementation.
> 
> The proposed solution centralizes this logic within `constants.py` and exposes a simple public function (`setup_fabric_fqdn`) that:
> 
> * Encapsulates the transformation logic in one place
> * Ensures consistent and correct FQDN generation
> * Provides a clean and discoverable API for users
> * Reduces duplication across projects and pipelines
> 
> From a design perspective, this moves the approach from a **manual workaround** to a **first-class, supported pattern** within the library, while still preserving backward compatibility.
> 
> Happy to adjust the approach if you’d prefer a lighter-touch solution (for example, keeping it internal-only or documenting the pattern more clearly). Would love to hear your thoughts.

Thanks for the explanation! Will review it and provide feedback.

src/fabric_cicd/constants.py
```diff
@@ -166,6 +166,74 @@ class OperationType(str, Enum):
...
+        >>> setup_fabric_fqdn(workspace_id)
+        >>>
+        >>> token_credential = AzureCliCredential()
+        >>> workspace = FabricWorkspace(
+        ...     workspace_id=workspace_id,
+        ...     repository_directory="/path/to/workspace",
+        ...     token_credential=token_credential
+        ... )
+    """
+    global FABRIC_API_ROOT_URL, DEFAULT_API_ROOT_URL
```
shirasassoon: 1. Input values should be validated when consuming them (workspace id, url) - leverage existing validation functions
2. Throw a warning if overwriting previously set URL value as a measure of protection if the function gets called more than once

```
def setup_fabric_fqdn(workspace_id: str) -> None:
    from fabric_cicd._common._validate_input import validate_workspace_id
    from fabric_cicd._common._validate_env_vars import validate_api_url

    global FABRIC_API_ROOT_URL, DEFAULT_API_ROOT_URL

    workspace_id = validate_workspace_id(workspace_id)  # raises on invalid input
    fqdn_url = _get_fabric_fqdn_url(workspace_id)
    fqdn_url = validate_api_url(fqdn_url, "setup_fabric_fqdn")  # safety net

    if FABRIC_API_ROOT_URL != "https://api.fabric.microsoft.com":
        logger.warning(
            f"setup_fabric_fqdn: overwriting previously set FABRIC_API_ROOT_URL '{FABRIC_API_ROOT_URL}'"
        )

    FABRIC_API_ROOT_URL = fqdn_url
    DEFAULT_API_ROOT_URL = fqdn_url
```

shirasassoon (commented):
Please add a changie, see https://github.com/microsoft/fabric-cicd/blob/main/CONTRIBUTING.md for instructions.

src/fabric_cicd/constants.py
```diff
@@ -166,6 +166,74 @@ class OperationType(str, Enum):
...
+    """
+    # Remove all dashes from workspace_id
+    no_dashes = workspace_id.replace("-", "")
+    # Extract first 2 characters
+    first_two = no_dashes[:2]
+    # Format as FQDN URL
+    return f"https://{no_dashes}.z{first_two}.w.api.fabric.microsoft.com"
+
+
+def setup_fabric_fqdn(workspace_id: str) -> None:
```
shirasassoon: Recommend renaming to `configure_fabric_fqdn` for clearer intent

shirasassoon (commented):
Users who hit this issue in production will likely get a connection error first and go to troubleshooting.md. Please add a short entry to the "Common Issues and Solutions" section there, something along these lines:

```
#### Private Link Connection Failures

**Symptom**: API calls fail with connection errors when deploying to a workspace
with "Allow connections only from workspace level private links" enabled.

**Solution**: Call `configure_fabric_fqdn` before initializing `FabricWorkspace`:

    from fabric_cicd import configure_fabric_fqdn, FabricWorkspace

    configure_fabric_fqdn(workspace_id)
    workspace = FabricWorkspace(workspace_id=workspace_id, ...)
```

shirasassoon (commented):
Please add a test file covering `_get_fabric_fqdn_url` (correct FQDN output for standard and dash-free inputs) and `configure_fabric_fqdn` (globals are correctly updated, overwrite warning fires on second call). Use `monkeypatch.setattr` to restore globals between tests to avoid cross-test pollution — this pattern is already used throughout the test suite.

src/fabric_cicd/constants.py
```diff
@@ -166,6 +166,74 @@ class OperationType(str, Enum):
...
+    """
+    Configure Fabric API URLs for private-link-enabled workspaces.
+
+    This function updates the global Fabric API URL constants to use the FQDN
```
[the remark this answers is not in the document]
SumeshKashyap: Test cases have been added in test_fqdn_workspace_id.py

src/fabric_cicd/constants.py
```diff
@@ -166,6 +166,74 @@ class OperationType(str, Enum):
...
+    format required for private-link-enabled workspaces. Call this function
+    before initializing a FabricWorkspace if you are using a private-link-enabled
+    workspace.
+
+    Args:
+        workspace_id: The workspace ID string, with or without dashes
+            (e.g., "f953f3da-c5f0-4e36-a644-c85933e35e2f" or "f953f3dac5f04e36a644c85933e35e2f").
```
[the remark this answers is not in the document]
SumeshKashyap: The new changes will only consider proper workspace GUID format with dashes. To avoid confusion, workspace id without dashes will not be supported by the the function

src/fabric_cicd/constants.py
```diff
@@ -166,6 +166,74 @@ class OperationType(str, Enum):
...
+        >>> workspace = FabricWorkspace(
+        ...     workspace_id=workspace_id,
+        ...     repository_directory="/path/to/workspace",
+        ...     token_credential=token_credential
+        ... )
+    """
+    global FABRIC_API_ROOT_URL, DEFAULT_API_ROOT_URL
+    fqdn_url = _get_fabric_fqdn_url(workspace_id)
+    FABRIC_API_ROOT_URL = fqdn_url
+    DEFAULT_API_ROOT_URL = fqdn_url
```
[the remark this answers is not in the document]
SumeshKashyap: Validation added in the function for workspace id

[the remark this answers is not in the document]
SumeshKashyap: Fixed this issue for a second call to different workspace by making FabricWorkspace read updated api url rather than old value in constants.DEFAULT_API_ROOT_URL.
image

src/fabric_cicd/constants.py
SumeshKashyap: Method name updated as suggested

src/fabric_cicd/constants.py
SumeshKashyap: Validation and overwritting measure protection added as suggested

src/fabric_cicd/constants.py
```diff
@@ -166,6 +168,88 @@ class OperationType(str, Enum):
 DEFAULT_API_ROOT_URL = validate_env_var_api_url(EnvVar.DEFAULT_API_ROOT_URL.value, "https://api.powerbi.com")
 FABRIC_API_ROOT_URL = validate_env_var_api_url(EnvVar.FABRIC_API_ROOT_URL.value, "https://api.fabric.microsoft.com")
 
+
+def _get_fabric_fqdn_url(workspace_id: str) -> str:
```
shirasassoon: Let's keep `constants.py` clean and move function logic to a different file. `_get_fabric_fqdn_url()` would be appropriate in `_common/_validate_env_vars.py`.

src/fabric_cicd/constants.py
```diff
@@ -166,6 +168,88 @@ class OperationType(str, Enum):
...
+    if not re.match(VALID_GUID_REGEX, workspace_id):
+        msg = f"workspace_id must be a valid GUID with dashes, got: '{workspace_id}'"
+        raise ValueError(msg)
+    no_dashes = workspace_id.replace("-", "")
+    first_two = no_dashes[:2]
+    return f"https://{no_dashes}.z{first_two}.w.api.fabric.microsoft.com"
+
+
+def configure_fabric_fqdn(workspace_id: str) -> None:
+    """
```
shirasassoon: Same comment for this logic - this one can be added to the `__init__.py` file since it is a public function

src/fabric_cicd/constants.py
```diff
@@ -166,6 +168,88 @@ class OperationType(str, Enum):
...
+    workspace_id = validate_workspace_id(workspace_id)  # raises on invalid input
+    fqdn_url = _get_fabric_fqdn_url(workspace_id)
+    fqdn_url = validate_api_url(fqdn_url, "configure_fabric_fqdn")  # safety net
+    
+    if FABRIC_API_ROOT_URL != "https://api.fabric.microsoft.com":
+        logger.warning(
+            f"configure_fabric_fqdn: overwriting previously set FABRIC_API_ROOT_URL '{FABRIC_API_ROOT_URL}'"
+        )
+
+    fqdn_url = _get_fabric_fqdn_url(workspace_id)
```
shirasassoon: This is a duplicate call and will negate the previous validation. This should be removed.

src/fabric_cicd/constants.py
```diff
@@ -3,7 +3,9 @@
 
 """Constants for the fabric-cicd package."""
 
+from asyncio.log import logger
```
shirasassoon: This is the incorrect logging import however, once the functions get moved to the respective files, this can be removed as well. The files should already have logging imported.

src/fabric_cicd/constants.py
```diff
@@ -166,6 +168,88 @@ class OperationType(str, Enum):
...
+    Configure Fabric API URLs for private-link-enabled workspaces.
+
+    This function updates the global Fabric API URL constants to use the FQDN
+    format required for private-link-enabled workspaces. Call this function
+    before initializing a FabricWorkspace if you are using a private-link-enabled
+    workspace.
+
+    Args:
+        workspace_id: The workspace ID string, with or without dashes
+            (e.g., "f953f3da-c5f0-4e36-a644-c85933e35e2f" or "f953f3dac5f04e36a644c85933e35e2f").
```
shirasassoon: Docstring needs to be updated since only dashed guids are supported.

shirasassoon (commented):
Tests look good, one gap to address is ensuring the snapshot works as expected. I would recommend adding a test in the `test_fabric_workspace.py` file. Feel free to reference this copilot suggested test:

```
def test_api_root_url_snapshot_is_not_retargeted_by_second_configure_call(
    temp_workspace_dir, patched_fabric_workspace, monkeypatch
):
    """Test that a constructed FabricWorkspace retains its snapshotted URL
    even if configure_fabric_fqdn is called again for a different workspace."""
    import fabric_cicd.constants as constants
    from fabric_cicd import configure_fabric_fqdn

    workspace_id_A = "f953f3da-c5f0-4e36-a644-c85933e35e2f"
    workspace_id_B = "a1b2c3d4-e5f6-7890-abcd-ef1234567890"

    # Reset globals to defaults before test
    monkeypatch.setattr(constants, "DEFAULT_API_ROOT_URL", "https://api.powerbi.com")
    monkeypatch.setattr(constants, "FABRIC_API_ROOT_URL", "https://api.fabric.microsoft.com")

    # Configure for workspace A and construct it
    configure_fabric_fqdn(workspace_id_A)
    expected_fqdn_A = constants.DEFAULT_API_ROOT_URL  # snapshot what was set

    with patch.object(FabricWorkspace, "_refresh_repository_items"):
        workspace_A = patched_fabric_workspace(
            workspace_id=workspace_id_A,
            repository_directory=str(temp_workspace_dir),
        )

    # Now configure for workspace B
    configure_fabric_fqdn(workspace_id_B)

    # workspace_A should still use FQDN_A, not FQDN_B
    assert workspace_A._api_root_url == expected_fqdn_A
    assert workspace_A.base_api_url.startswith(expected_fqdn_A)

```

src/fabric_cicd/constants.py
SumeshKashyap: chnages  made as per suggestions but please note that I introduced a new _VALID_GUID_REGEX in _validate_env_vars.py because moving _get_fabric_fqdn_url there prevented importing VALID_GUID_REGEX from constants.py due to circular import issues.
A cleaner solution might be  to move VALID_GUID_REGEX to _validate_env_vars.py and re-export it from constants.py, so all modules can reuse it without circular dependencies. Please let me know your suggestion if you prefer the way it is done or we move VALID_GUID_REGEX to _validate_env_vars.py? thanks

src/fabric_cicd/constants.py
SumeshKashyap: configure_fabric_fqdn has been added to __init__.py

src/fabric_cicd/constants.py
SumeshKashyap: this has been removed now

src/fabric_cicd/constants.py
SumeshKashyap: The docstring has been updated

src/fabric_cicd/constants.py
shirasassoon: @SumeshKashyap I think we can go with the cleaner solution.

shirasassoon:
@SumeshKashyap All the updates looks good. Please address the linting failure and the REGEX constant. Once you push the last changes, we should be ready to merge.

src/fabric_cicd/constants.py
SumeshKashyap: clean solution applied

SumeshKashyap:
> @SumeshKashyap All the updates looks good. Please address the linting failure and the REGEX constant. Once you push the last changes, we should be ready to merge.
@shirasassoon I have made  the discussed changes in the last commit

src/fabric_cicd/constants.py
```diff
@@ -6,7 +6,7 @@
 import os
 from enum import Enum
 
-from fabric_cicd._common._validate_env_vars import validate_env_var_api_url
+from fabric_cicd._common._validate_env_vars import VALID_GUID_REGEX, validate_env_var_api_url  # noqa: F401
```
shirasassoon, line 11: Alternatively, you can try the following to avoid the linting issue 

`from fabric_cicd._common._validate_env_vars import VALID_GUID_REGEX as VALID_GUID_REGEX`

src/fabric_cicd/constants.py
```diff
@@ -166,6 +166,7 @@ class OperationType(str, Enum):
 DEFAULT_API_ROOT_URL = validate_env_var_api_url(EnvVar.DEFAULT_API_ROOT_URL.value, "https://api.powerbi.com")
 FABRIC_API_ROOT_URL = validate_env_var_api_url(EnvVar.FABRIC_API_ROOT_URL.value, "https://api.fabric.microsoft.com")
 
+
```
shirasassoon: nit - remove this extra line

src/fabric_cicd/constants.py
SumeshKashyap: Change made as suggested and linting issue resolved

src/fabric_cicd/constants.py
SumeshKashyap: extra line removed

shirasassoon (approved):

PR #960, opened 2026-04-28
fix(_check_utils): add 2s HTTP timeout to startup version + changelog fetches

SAY-5:
Reported in microsoft/fabric-cli#224 — every `fab` command was hanging
indefinitely at startup because import-time `fabric_cicd` work calls
out to pypi.org with no timeout.

## Why

`fabric_cicd/__init__.py` calls `check_version()`, which does:

```python
response = requests.get("https://pypi.org/pypi/fabric-cicd/json")
```

with no `timeout=` set, so the OS socket default applies. Under poor
network conditions (corporate proxies, broken IPv6 paths, slow DNS,
transient pypi outages) this stalls for minutes-to-hours and never
returns. The sister `parse_changelog()` request to
`raw.githubusercontent.com` had the same problem.

Both call sites are already wrapped in `try/except Exception`, so all
that's missing is the timeout — once that's set, a slow upstream just
becomes a quiet debug log and CLI startup proceeds normally.

## Fix

- New module-level constant `_HTTP_TIMEOUT_SECONDS = 2`.
- Pass it as `timeout=_HTTP_TIMEOUT_SECONDS` to both `requests.get(...)`
  calls in `_check_utils.py`.

This is the minimum possible change. The `fabric-cli` issue also
suggests moving the version check off the import path entirely
(deferred / async / TTY-gated) — that's a bigger restructure and a
separate PR if you'd like one.

## Verification

- `python -m py_compile src/fabric_cicd/_common/_check_utils.py` →
  clean.
- I confirmed the only ungated `requests.get` in this file is the
  changelog fetch already covered.

### DIFF
--- a/src/fabric_cicd/_common/_check_utils.py
+++ b/src/fabric_cicd/_common/_check_utils.py
@@ -19,6 +19,13 @@ from fabric_cicd._common._exceptions import FileTypeError
 
 logger = logging.getLogger(__name__)
 
+# Cap the upstream HTTP calls (changelog fetch, pypi version check) so a
+# slow/unreachable host can't block CLI startup. The previous code passed
+# no `timeout=` and inherited the OS socket default, which can stall for
+# minutes-to-hours under poor network conditions (corporate proxies, broken
+# IPv6 paths, transient pypi outages). See microsoft/fabric-cli#224.
+_HTTP_TIMEOUT_SECONDS = 2
+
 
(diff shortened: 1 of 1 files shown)


shirasassoon:
@SAY-5 Thanks for raising a PR to address this known issue. We are planning to remove the version check entirely to prevent unnecessary noise and slowdowns during import so this PR will not be needed.

PR #972, opened 2026-05-05
Closes #973 Remove version check on package import

### LINKED ISSUE #973: [TECH DEBT] Remove version check on package import
### Problem

The `fabric-cicd` package currently makes a network call to PyPI on every import to check if a newer version is available. This causes issues for applications built on top of `fabric-cicd` that intentionally pin versions:

- Generates noisy upgrade warnings that are irrelevant when versions are deliberately pinned
- Adds latency on import due to the PyPI network call
- Fails or times out in air-gapped/firewalled environments
- Libraries should not make network calls as an import side effect

### Proposed change

Remove the `check_version()` functionality entirely rather than keeping it opt-out, since:

- The primary audience (CI/CD pipelines) pins dependencies and may lack outbound network access
- `pip` and package managers (dependabot, renovate) already handle version notifications better

### Scope

- Remove `check_version()` and `parse_changelog()` from `_check_utils.py`
- Remove the import and invocation in `__init__.py`
- Remove `VERSION_CHECK_DISABLED` from `EnvVar` enum and `constants.py`
- Remove `packaging` as a runtime dependency

shirasassoon:
This pull request removes the version check feature and its related dependencies from the codebase. The `packaging` library is no longer required, and all code related to checking for new releases and displaying changelog information at startup has been deleted. This simplifies the project and reduces unnecessary runtime dependencies.

Key changes:

**Dependency removal:**
- Removed the `packaging` library from both the runtime dependencies in `.github/copilot-instructions.md` and the `pyproject.toml` file, as it is no longer needed. [[1]](diffhunk://#diff-227c2c26cb2ee0ce0f46a320fc48fbcbdf21801a57f59161b1d0861e8aad55f5L176-R176) [[2]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L23)

**Feature removal – Version check:**
- Deleted the version check logic, including the `check_version` and `parse_changelog` functions, from `src/fabric_cicd/_common/_check_utils.py`. This also removes the use of `requests`, `packaging`, and related constants and color formatting.
- Removed the import and invocation of `check_version` from `src/fabric_cicd/__init__.py`, so the version check is no longer performed at startup. [[1]](diffhunk://#diff-364c25e03dde1816a460702e5c0224b29c1bd8d185d9eea1089520dd040c9849L10) [[2]](diffhunk://#diff-364c25e03dde1816a460702e5c0224b29c1bd8d185d9eea1089520dd040c9849L138-L140)
- Eliminated the `VERSION_CHECK_DISABLED` environment variable and associated logic from `src/fabric_cicd/constants.py`, as it is no longer relevant. [[1]](diffhunk://#diff-91c2d32be351f155f090c4bd57ebe3c10e646edbc1b663e2a72b697f956a949cL40-L41) [[2]](diffhunk://#diff-91c2d32be351f155f090c4bd57ebe3c10e646edbc1b663e2a72b697f956a949cL183-L185)


uv.lock
```diff
@@ -1,5 +1,4 @@
 version = 1
-revision = 3
```
aviatco, line 2: does this file should be checked in?

src/fabric_cicd/__init__.py
```diff
@@ -135,9 +134,6 @@ def disable_file_logging() -> None:
 configure_logger()
 sys.excepthook = exception_handler
 
-if not constants.VERSION_CHECK_DISABLED:
```
aviatco, line 138: so you decided to remove the check entirely? not to changed the env var to enabled?

shirasassoon: Yes - see the reasoning described in the issue #973

uv.lock
shirasassoon: This is auto-generated. I removed the `packaging` dependency from `pyproject.toml` then ran `uv sync --dev` which made these updates.

aviatco: ok, just usually .lock files should be in gitIgnore

aviatco (approved):

PR #990, opened 2026-05-14
Refactor _refresh_token() for clarity

shirasassoon:
This pull request refactors the authentication token management in `FabricEndpoint` and significantly expands the test coverage for token handling and retry logic. The main improvements include switching to a cached token approach with proactive refresh, updating how token expiry and retries are handled, and adding comprehensive tests for token caching, expiry, retry logic, and tracer usage.

**Authentication and token management:**

- Refactored the token management in `FabricEndpoint` to use a cached token (`self._token`) and expiry (`self._token_expiry`) with a 10-second buffer, replacing the previous eager refresh logic. The token is now validated and cached at initialization, and refreshed only when needed. (`src/fabric_cicd/_common/_fabric_endpoint.py` [[1]](diffhunk://#diff-df48b77e163b4a13c97f80408ad6640fca48ffdc60c19e085b4c73589933178cL43-R50) [[2]](diffhunk://#diff-df48b77e163b4a13c97f80408ad6640fca48ffdc60c19e085b4c73589933178cL150-R169)
- Updated the `invoke` method to always use the current token via `_get_token()`, and to invalidate the cache and retry if a `401 TokenExpired` response is received, improving reliability and reducing unnecessary token requests. (`src/fabric_cicd/_common/_fabric_endpoint.py` [[1]](diffhunk://#diff-df48b77e163b4a13c97f80408ad6640fca48ffdc60c19e085b4c73589933178cL82-R83) [[2]](diffhunk://#diff-df48b77e163b4a13c97f80408ad6640fca48ffdc60c19e085b4c73589933178cL98-R100)

**Testing improvements:**

- Added new tests for token caching, expiry buffer logic, and proactive refresh, ensuring tokens are only refreshed when necessary and that the buffer is respected. (`tests/test__fabric_endpoint.py` [tests/test__fabric_endpoint.pyL117-R178](diffhunk://#diff-d5975d64f0adde0e560b6984da9e6c82903b4d6c09da96d60e3051c37465497aL117-R178))
- Added tests to verify that the token is correctly sent in the `Authorization` header, and that exceptions during token acquisition are properly handled and wrapped. (`tests/test__fabric_endpoint.py` [[1]](diffhunk://#diff-d5975d64f0adde0e560b6984da9e6c82903b4d6c09da96d60e3051c37465497aL117-R178) [[2]](diffhunk://#diff-d5975d64f0adde0e560b6984da9e6c82903b4d6c09da96d60e3051c37465497aL244-R346)
- Expanded tests for retry logic, including connection errors, timeouts, exponential backoff, `Retry-After` header handling, environment variable overrides, and max duration enforcement. (`tests/test__fabric_endpoint.py` [[1]](diffhunk://#diff-d5975d64f0adde0e560b6984da9e6c82903b4d6c09da96d60e3051c37465497aL191-R334) [[2]](diffhunk://#diff-d5975d64f0adde0e560b6984da9e6c82903b4d6c09da96d60e3051c37465497aR574-R636)
- Added tests to ensure the HTTP tracer is called as expected during requests. (`tests/test__fabric_endpoint.py` [tests/test__fabric_endpoint.pyL191-R334](diffhunk://#diff-d5975d64f0adde0e560b6984da9e6c82903b4d6c09da96d60e3051c37465497aL191-R334))

**General test code cleanup:**

- Removed obsolete or redundant tests and updated test names and assertions to reflect the refactored token logic and improved error messages. (`tests/test__fabric_endpoint.py` [[1]](diffhunk://#diff-d5975d64f0adde0e560b6984da9e6c82903b4d6c09da96d60e3051c37465497aL55-L84) [[2]](diffhunk://#diff-d5975d64f0adde0e560b6984da9e6c82903b4d6c09da96d60e3051c37465497aL244-R346)

These changes collectively improve the reliability, efficiency, and testability of authentication and retry handling in the FabricEndpoint class.


src/fabric_cicd/_common/_fabric_endpoint.py
```diff
@@ -147,27 +148,30 @@ def invoke(
...
+        """Gets an AAD token, using a cached value if still valid."""
+        resource_url = "https://api.fabric.microsoft.com/.default"
+
         if (
-            self.aad_token is None
-            or self.aad_token_expiration is None
-            or self.aad_token_expiration < datetime.datetime.now(datetime.timezone.utc)
+            self._token is not None
+            and self._token_expiry is not None
+            and self._token_expiry > datetime.datetime.now(datetime.timezone.utc)
```
tarostok: should we add some "buffer" for expiry time such that if the token expires in less than 10 seconds we will get a new one?

shirasassoon: We can add the buffer for the following reasons:

- Avoids race conditions where a token is valid at check time but expired by the time the server processes the request
- Eliminates unnecessary 401 → retry round-trips (saves ~1 network call + latency)

Having a buffer will result in slightly more frequent token refreshes but since it is only 10 sec it will add negligible overhead so no concern from that end.

src/fabric_cicd/_common/_fabric_endpoint.py
```diff
@@ -147,27 +148,31 @@ def invoke(
             "status_code": response.status_code,
         }
 
-    def _refresh_token(self) -> None:
-        """Refreshes the AAD token if empty or expiration has passed."""
+    def _get_token(self) -> str:
+        """Gets an AAD token, using a cached value if still valid."""
+        resource_url = "https://api.fabric.microsoft.com/.default"
+        token_expiry_buffer = datetime.timedelta(seconds=10)
```
ayeshurun: we can set it once at the top of this module instead initializing every method call, like a constant

src/fabric_cicd/_common/_fabric_endpoint.py
```diff
@@ -147,27 +148,31 @@ def invoke(
...
+
+        try:
+            access_token = self.token_credential.get_token(resource_url)
+            self._token = access_token.token
+            self._token_expiry = datetime.datetime.fromtimestamp(
+                access_token.expires_on, tz=datetime.timezone.utc
+            )
+            return self._token
+        except ClientAuthenticationError as e:
+            msg = f"Failed to acquire AAD token. {e}"
```
ayeshurun: AAD is the legacy name. the official name is Entra token or Microsoft Entra Token

src/fabric_cicd/_common/_fabric_endpoint.py
```diff
@@ -147,27 +148,31 @@ def invoke(
...
+            raise TokenError(msg, logger) from e
+        except Exception as e:
+            msg = f"An unexpected error occurred when generating the AAD token. {e}"
```
ayeshurun: same here regarding AAD

src/fabric_cicd/_common/_fabric_endpoint.py
```diff
@@ -95,8 +96,8 @@ def invoke(
 
                 # Handle expired authentication token
                 if response.status_code == 401 and response.headers.get("x-ms-public-api-error-code") == "TokenExpired":
-                    logger.info(f"{constants.INDENT}AAD token expired. Refreshing token.")
-                    self._refresh_token()
+                    self._token = None  # Invalidate cache to force refresh
+                    logger.info(f"{constants.INDENT}AAD token expired. Retrying with refreshed token.")
```
ayeshurun: same comment regarding AAD

tests/test_integration_token.py
```diff
@@ -0,0 +1,127 @@
...
+
+These tests require an authenticated Azure CLI session (`az login`).
+They are skipped by default unless opted in via env var.
+
+Usage:
+    FABRIC_CICD_LIVE_TOKEN_TESTS=1 uv run pytest tests/test_integration_token.py -v
+    # PowerShell:
+    $env:FABRIC_CICD_LIVE_TOKEN_TESTS="1"; uv run pytest tests/test_integration_token.py -v
+"""
+
```
ayeshurun, line 14: can you elaborate? why this was added now?

shirasassoon: Initially we were going to remove the expiration handling and token caching, however, when running this test, it revealed a regression where Azure CLI auth method was failing without caching. The test isn't technically needed or run in CI (unless env vars are set), but it is more of a future proof in case someone decides to refactor the authentication flow.

tarostok (approved):

ayeshurun (approved):

PR #1012, opened 2026-06-04
Closes #1020 Add new deployment mode (bulk publish)

### LINKED ISSUE #1020: [FEATURE] Add bulk publish support using Fabric bulk import API
### Use Case / Problem

Currently, fabric-cicd publishes items one at a time in a predefined dependency order (serial mode) which requires explicit dependency ordering. The Fabric platform now offers a [bulk import API](https://learn.microsoft.com/en-us/rest/api/fabric/core/items/bulk-import-item-definitions(beta)?tabs=HTTP) (beta) that accepts all items in a single call and resolves dependencies automatically via a server-side dependency graph using logical IDs.

This feature should:

- Allow users to deploy all items in a single API call via `POST /items/bulkImportDefinitions?beta=True`

- Leverage the Fabric dependency graph to eliminate client-side ordering

- Be gated behind experimental feature flags (`enable_experimental_features` + `enable_bulk_publish`) until the API becomes GA

- Gracefully fall back to standard serial mode when incompatible configurations are detected

### Proposed Solution

1. **Turn on using feature flags**: Add `enable_bulk_publish` and require `enable_experimental_features` 
2. **Bulk publish flow** 
   - Check for unsupported types - fallback to standard publish when present
   - Check if the parameter file contains dynamic replacement variables (`$workspace.*`, `$items.*`) — these require runtime resolution and are incompatible with bulk mode. Fallback to standard publish when present
   - Selective deployment features are not supported in phase 1
   - Call bulk publisher 
   - Call async post processing
  
3. **Bulk API integration**
   - Add separate publish method that prepares the payload definition parts and calls the bulk import API
   - Apply item type-specific file processing (when applicable) and support parameterization (except dynamic variables)

4. **Bulk orchestration** (in `_base_publisher.py`):
   - Add `publish_all_bulk()` static method that runs `pre_publish_all()` per type, collects all items into a single pool, makes one bulk API call, then runs `post_publish_all()` per type.

### Alternatives Considered

_No response_

### Impact Assessment

- [ ] This would help me personally
- [ ] This would help my team/organization
- [x] This aligns with Microsoft Fabric roadmap items

### Implementation Notes

_No response_

shirasassoon:
This pull request introduces a major new experimental feature: bulk publish mode, which allows deploying all items in a single API call using the Fabric bulk import API. The documentation is significantly expanded to explain this feature, its usage, limitations, and troubleshooting. Additionally, there are improvements to error reporting and minor code cleanups. Below are the most important changes:

**Bulk Publish Feature & Documentation:**

* Added support for bulk publish mode, enabling deployment of all items in a single API call via the Fabric bulk import API. This is marked as experimental and requires enabling both `enable_experimental_features` and `enable_bulk_publish` flags. Documentation includes usage examples (programmatic and config-based), current limitations, and guidance for logical ID handling. [[1]](diffhunk://#diff-fe18817e34a8905d56f3b7ed9c37c1adeb23577736c22c4d53bbbaf755cff55fR1-R8) [[2]](diffhunk://#diff-c6d5a0fcefe940477d9234afcbed07e3a42c359ef09ebf0093905ea3c9c7f8f1L7-R51) [[3]](diffhunk://#diff-c6d5a0fcefe940477d9234afcbed07e3a42c359ef09ebf0093905ea3c9c7f8f1R63-R152)
* Updated troubleshooting documentation with a new section on common bulk publish failures and their solutions, including dependency resolution, unsupported item types, and error message interpretation.

**Feature Flags & Selective Deployment:**

* Restructured the feature flags documentation to group flags by theme (Publish behavior, Unpublish behavior, Selective deployment, Diagnostics). Clarified which features are experimental and improved explanations for selective deployment, including compatibility with bulk publish. [[1]](diffhunk://#diff-c6d5a0fcefe940477d9234afcbed07e3a42c359ef09ebf0093905ea3c9c7f8f1L7-R51) [[2]](diffhunk://#diff-c6d5a0fcefe940477d9234afcbed07e3a42c359ef09ebf0093905ea3c9c7f8f1R63-R152)
* Clarified the behavior of folder-based filters, noting that root-level items are not impacted, and improved explanations of filter precedence and mutual exclusivity. [[1]](diffhunk://#diff-c6d5a0fcefe940477d9234afcbed07e3a42c359ef09ebf0093905ea3c9c7f8f1L62-R174) [[2]](diffhunk://#diff-c6d5a0fcefe940477d9234afcbed07e3a42c359ef09ebf0093905ea3c9c7f8f1L89-R201)

**Parameterization Limitations:**

* Added documentation note that dynamic replacement variables (`$workspace`, `$items`) are not supported in bulk publish mode; deployments with these variables automatically fall back to standard publish.

**Error Handling Improvements:**

* Enhanced error reporting in `_handle_response` to extract and display more detailed error messages from the Fabric API, including related resources and error codes, for better troubleshooting of failed operations.


src/fabric_cicd/fabric_workspace.py
```diff
@@ -810,6 +780,93 @@ def _publish_item(
...
+        path_prefix = f"{folder_path}/{item_dir_name}"
+
+        parts = []
+        for file in item.item_files:
+            if re.match(exclude_path, file.relative_path):
+                continue
+            if file.type == "text" and not str(file.file_path).endswith(".platform"):
+                file.contents = func_process_file(self, item, file) if func_process_file else file.contents
+                file.contents = self._replace_parameters(file, item)
+
```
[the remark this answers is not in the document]
shirasassoon: The bulk API resolves workspace IDs and logical IDs server-side, so those client-side replacements are unnecessary (and would be redundant). The feedback is not valid.

src/fabric_cicd/fabric_workspace.py
```diff
@@ -810,6 +777,105 @@ def _publish_item(
             logger.info(f"{constants.INDENT}Published {item_type} '{item_name}'")
         return
 
+    def _publish_items(
```
ohadedry, line 780: function name doesn't indicate that this is bulk import, if relevant then consider renaming.

src/fabric_cicd/fabric_workspace.py
```diff
@@ -810,6 +777,105 @@ def _publish_item(
...
+
+            # Assign GUIDs from the response
+            if (item_id and item_type in self.repository_items) and (item_name in self.repository_items[item_type]):
+                self.repository_items[item_type][item_name].guid = item_id
+
+            # Store response if responses are being tracked
+            if self.responses is not None:
+                if item_type not in self.responses:
+                    self.responses[item_type] = {}
+                self.responses[item_type][item_name] = {
```
ohadedry, line 827: There is single API call with single response, why do you keep it per item?

src/fabric_cicd/_items/_base_publisher.py
```diff
@@ -298,6 +299,74 @@ def get_orphaned_items(
...
+        items_with_context: list[tuple[str, Item, ItemPublisher]] = []
+        skipped_items: list[str] = []
+
+        # Phase 1: Pre-hooks + collect items with their publisher context
+        for _order_num, item_type in ItemPublisher.get_item_types_to_publish(fabric_workspace_obj):
+            publisher = ItemPublisher.create(item_type, fabric_workspace_obj)
+            type_items = publisher.get_items_to_publish()
+
+            # Filter 1: items_to_include — mark items not in the include list as skip_publish=True
+            all_type_items = fabric_workspace_obj.repository_items.get(item_type.value, {})
```
aviatco: correct me if im wrong, but dont  we have those filters and flow exists already (regarding most of the logic in this function)? if so, maybe you can extract the share logic and reuse it

src/fabric_cicd/_items/_lakehouse.py
```diff
@@ -192,12 +195,18 @@ def publish_one(self, _shortcut_name: str, shortcut: dict) -> None:
             msg = f"Failed to publish '{shortcut['name']}' for lakehouse {self.item_obj.name}"
             raise FailedPublishedItemStatusError(msg, logger) from e
 
-    def publish_all(self) -> None:
+    def publish_all(self, log_header_flag: bool = True) -> bool:
```
ohadedry: Publisher abstract method is defined to return None but here you changed it to bool. Which one is the correct one?

src/fabric_cicd/_items/_environment.py
```diff
@@ -162,6 +163,10 @@ def _check_environment_publish_state(fabric_workspace_obj: FabricWorkspace, init
         )
     ]
 
+    if not filtered_environments:
```
aviatco, line 166: is this the environment user pass in the config file? now this is mandatory?

src/fabric_cicd/_items/_lakehouse.py
```diff
@@ -130,11 +131,13 @@ def publish_one(self, item_name: str, item: Item) -> None:
     def post_publish_all(self) -> None:
         """Publish shortcuts after all lakehouses are published to protect interrelationships."""
         if FeatureFlag.ENABLE_SHORTCUT_PUBLISH.value in constants.FEATURE_FLAG:
+            header_logged = False
             for item_obj in self.fabric_workspace_obj.repository_items.get(self.item_type, {}).values():
                 # Check if the item is published to avoid any post publish actions
                 if not item_obj.skip_publish and item_obj.guid:
                     shortcut_publisher = ShortcutPublisher(self.fabric_workspace_obj, item_obj)
-                    shortcut_publisher.publish_all()
+                    if shortcut_publisher.publish_all(log_header_flag=not header_logged):
```
aviatco: nit - you can do:
header_logged = shortcut_publisher.publish_all(log_header_flag=not header_logged):

src/fabric_cicd/publish.py
```diff
@@ -247,15 +271,21 @@ def publish_all_items(
...
+    if fabric_workspace_obj.bulk_publish_enabled:
+        # Publish all items in bulk (experimental)
+        log_header(logger, "Publishing Items in Bulk")
+        publishers_with_async_check = items.ItemPublisher.publish_all_bulk(fabric_workspace_obj)
+    else:
+        # Publish items in the defined order synchronously (standard)
+        total_item_types = len(constants.SERIAL_ITEM_PUBLISH_ORDER)
+        publishers_with_async_check: list[items.ItemPublisher] = []
+        for order_num, item_type in items.ItemPublisher.get_item_types_to_publish(fabric_workspace_obj):
+            log_header(logger, f"Publishing Item {order_num}/{total_item_types}: {item_type.value}")
```
aviatco, line 284: so this will be printed for each item right?

src/fabric_cicd/_items/_environment.py
```diff
@@ -234,6 +244,7 @@ class EnvironmentPublisher(ItemPublisher):
 
     item_type = ItemType.ENVIRONMENT.value
     has_async_publish_check = True
+    func_process_file = staticmethod(_process_environment_file)
```
ohadedry, line 247: Can you please explain this line?

src/fabric_cicd/_items/_variablelibrary.py
```diff
@@ -56,3 +57,14 @@ def publish_one(self, item_name: str, item: Item) -> None:
...
             activate_value_set(self.fabric_workspace_obj, item)
+
+    def post_publish_all(self) -> None:
+        """Activate value sets after bulk upload."""
+        if self.fabric_workspace_obj.bulk_publish_enabled:
+            header_logged = False
+            for item in self.fabric_workspace_obj.repository_items.get(self.item_type, {}).values():
+                if not item.skip_publish:
+                    if not header_logged:
+                        log_header(logger, "Activating Variable Library Value Sets")
```
ohadedry: Could you log the header before iterating the items and avoid the header_logged flag? (similar to how it is done in _environment.py)

ohadedry (commented):
Left some comments, still need to review tests

tests/test_shortcut_exclude.py
```diff
@@ -413,7 +413,7 @@ def test_publish_all_marks_excluded_items_skip_publish():
     # Track which items get shortcut-published
     shortcut_published_guids = []
 
-    def fake_shortcut_publish_all(self_inner):
+    def fake_shortcut_publish_all(self_inner, **_kwargs):
```
ohadedry: Why the new property is required?

src/fabric_cicd/_items/_environment.py
shirasassoon: No this is specifically referring to an Environment item - this is a minor optimization - we do an early return if there are no Environment items to check - to avoid printing this line 
`"Checking Environment Publish State for [ ]"` as well as avoiding making an unnecessary API call

src/fabric_cicd/publish.py
shirasassoon: This logic is the same as it was before - it prints the header for each item type.

src/fabric_cicd/_items/_lakehouse.py
shirasassoon: Fixed

src/fabric_cicd/fabric_workspace.py
shirasassoon: I hear your feedback; I would keep as is since it's an internal method which gets called by the `publish_all_bulk()` so the association is there in addition to the method's docstring - the goal is to have one `publish_items()` method eventually.

src/fabric_cicd/_items/_base_publisher.py
shirasassoon: Fixed

tests/test_bulk_publish.py
```diff
@@ -0,0 +1,933 @@
...
+        },
+        "config": {"logicalId": logical_id},
+    }
+    with platform_file.open("w", encoding="utf-8") as f:
+        json.dump(metadata, f)
+
+    with (item_dir / "dummy.txt").open("w", encoding="utf-8") as f:
+        f.write("Dummy file content")
+
+    return item_dir
```
ohadedry, line 103: Minor: It's not intuitive that create_test_item returns directory and not item

src/fabric_cicd/fabric_workspace.py
shirasassoon: I agree the header and status are redundant in the bulk case since it is a single API call. This was done to make it consistent with the single API response collection since we are using the same dict and therefore the structure should be the same. I can try to think of optimizing it for the bulk scenario.

src/fabric_cicd/_items/_environment.py
shirasassoon: This class attribute is used exclusively by the bulk publish path to apply custom file processing for specific item types (like Environment) before converting definition files to base64. In the single-item path, each publisher's `publish_one` method calls `_publish_item()` directly and passes `func_process_file` as a parameter. The bulk path, however, bypasses individual publisher methods entirely - it processes all items generically in `_prepare_bulk_item_parts`. This class attribute provides an effective way for the bulk path to discover and access each publisher's custom file processing logic without calling `publish_one`. It is wrapped in `staticmethod` to prevent Python from injecting the `publisher` instance as an extra first argument when the attribute is accessed via the publisher object which would break the expected signature.

src/fabric_cicd/_items/_variablelibrary.py
shirasassoon: Fixed

src/fabric_cicd/_items/_base_publisher.py
shirasassoon: Changed the approach so this code is removed

src/fabric_cicd/_items/_lakehouse.py
shirasassoon: Fixed the approach

tests/test_shortcut_exclude.py
shirasassoon: No longer needed after modifying approach for shortcut publishing header. Will remove it.

tests/test_bulk_publish.py
shirasassoon: Renamed to be clearer

tests/test_bulk_publish.py
```diff
@@ -0,0 +1,885 @@
...
+            assert len(publish_calls) == 1
+
+
+# =============================================================================
+# Bulk Publish Post-Publish Hook Tests
+# =============================================================================
+
+
+@pytest.mark.usefixtures("bulk_publish_flags")
+class TestBulkPublishPostPublishHooks:
```
ohadedry, line 475: As discussed offline, consider moving these tests to a new test file as they are PostPublish related and not specifically to bulk publish

ohadedry (approved):

tests/test_bulk_publish.py
shirasassoon: Noted. Will open a separate PR for test refactoring.

PR #1023, opened 2026-06-14
Tighten http trace security

shirasassoon:
This pull request enhances the security and robustness of the HTTP tracing functionality by improving how sensitive information is handled and how trace files are managed. The main changes include stricter filtering of sensitive HTTP headers from traces, validation of the trace file path, and improved user warnings about the risks of sharing trace files.

**Security and sensitive data handling:**

* Introduced `SENSITIVE_REQUEST_HEADERS` and `SENSITIVE_RESPONSE_HEADERS` constants in `constants.py` to define which HTTP headers should be excluded from trace logs, replacing the previous single `AUTHORIZATION_HEADER` filter. This broadens the protection to cover more sensitive headers in both requests and responses.
* Updated `capture_request` and `capture_response` methods in `_http_tracer.py` to use these new header sets, ensuring filtered headers are not written to trace files. [[1]](diffhunk://#diff-b140b68b21156024ab6ba5d6cf6f724032db1602fb3cd09034215caeb51fc724L171-R207) [[2]](diffhunk://#diff-b140b68b21156024ab6ba5d6cf6f724032db1602fb3cd09034215caeb51fc724L198-R234)

**Trace file management and user safety:**

* Added a `_validate_output_path` method to `_http_tracer.py` that enforces the trace file must have a `.json` extension and reside within the current working directory, preventing accidental path traversal or misnaming.
* Added a warning log in the tracer initialization to alert users that trace files may contain sensitive data and should not be committed or shared.

**Imports and dependencies:**

* Updated imports in `_http_tracer.py` to use the new sensitive header constants.


src/fabric_cicd/_common/_http_tracer.py
```diff
@@ -147,12 +147,45 @@ def __init__(self, output_file: Optional[str] = None) -> None:
...
+            raw_path: The raw file path string to validate.
+
+        Returns:
+            The resolved absolute path as a string.
+
+        Raises:
+            ValueError: If the path fails validation.
+        """
+        resolved = Path(raw_path).resolve()
+        cwd = Path.cwd().resolve()
```
ayeshurun: it might throw an error in case it couldnt resolve. consider try-catch blocks

shirasassoon: Fixed

src/fabric_cicd/_common/_http_tracer.py
```diff
@@ -147,12 +147,49 @@ def __init__(self, output_file: Optional[str] = None) -> None:
...
+
+        Returns:
+            The resolved absolute path as a string.
+
+        Raises:
+            ValueError: If the path fails validation.
+        """
+        try:
+            resolved = Path(raw_path).resolve()
+            cwd = Path.cwd().resolve()
```
aviatco, line 179: why we need cwd? is it used? or because we just have to resolve it?

shirasassoon: It is used to check if the file path exists within the current working directory (inside or equal to see line 188)

aviatco (approved):

PR #1038, opened 2026-06-23
Closes #1040: Enforce owner-only file permissions on library generated files on POSIX systems

### LINKED ISSUE #1040: [TECH DEBT] Enforce owner permissions for POSIX on fabric-cicd files with sensitive content
Tighten security through enforcing owner permissions for `http_trace.json` and `fabric_cicd.error.log` files which may sensitive information (API payloads) - this is only applied for Mac/Linux OS not Windows.

shirasassoon:
#1040
This pull request introduces a new utility for enforcing owner-only file permissions on POSIX systems and integrates it into file output and logging mechanisms to enhance security. It also adds comprehensive tests for this functionality.

**Security & Permissions Enforcement:**

* Added `restrict_file` function in `src/fabric_cicd/_common/_secure_io.py` to set owner-only (0o600) permissions on files, with no-ops on Windows and robust error handling.
* Updated `_flush_traces_to_file` in `src/fabric_cicd/_common/_http_tracer.py` to call `restrict_file` after writing trace output files, ensuring traces are not world-readable.

**Logging Improvements:**

* Introduced `_RestrictedFileHandler` in `src/fabric_cicd/_common/_logging.py`, a custom `FileHandler` that enforces owner-only permissions on log files after each emit, and updated the default file handler to use it.

**Testing:**

* Added `tests/test_secure_io.py` with tests for `restrict_file`, covering permission tightening, idempotency, missing files, and Windows no-op behavior.


src/fabric_cicd/_common/_http_tracer.py
```diff
@@ -242,7 +242,14 @@ def _flush_traces_to_file(self) -> None:
...
         }
 
-        with output_path.open("w") as f:
+        from fabric_cicd._common._secure_io import IS_POSIX, restrict_file, restricted_opener
+
+        # Tighten pre-existing files before writing (so truncation doesn't
+        # expose content through a previously world-readable fd).
+        restrict_file(self.output_file)
+
+        opener = restricted_opener if IS_POSIX else None
```
ayeshurun: is it limited to POSIX? i mean what will happen in Windows machines if we remove the code that checks the posix condition?

shirasassoon: Removing the condition would be harmless, but passing the mode would be ignored on Windows.

shirasassoon: We can remove the condition (implicit) or keep it (explicit) either way works.

ayeshurun: if its ignored on Windows, lets keep it implicit

shirasassoon: Updated to implicit

src/fabric_cicd/_common/_secure_io.py
```diff
@@ -0,0 +1,43 @@
...
+import os
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+# True on Linux/macOS; False on Windows where POSIX permission bits are a no-op.
+IS_POSIX = os.name != "nt"
+
+# Owner read/write only (rw-------).
+OWNER_ONLY_FILE_MODE = 0o600
```
aviatco, line 16: do we need to support restrict also folders? OWNER_ONLY_DIR_MODE = 0o700?

src/fabric_cicd/_common/_secure_io.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Intended for use as the ``opener`` argument to :func:`open` so that
+    new files are created with restricted permissions from the start,
+    eliminating any race window.
+    """
+    return os.open(path, flags, OWNER_ONLY_FILE_MODE)
+
+
+def restrict_file(path: str) -> None:
+    """Set owner-only permissions (0o600) on an existing file.
```
aviatco, line 31: What about existing files that were created without the restricted permission? do we have those kinds of files that need to be handled?

src/fabric_cicd/_common/_secure_io.py
shirasassoon: No there are no directories created by fabric-cicd, the two files get created in CWD

src/fabric_cicd/_common/_secure_io.py
shirasassoon: I believe there is tightening of existing files

aviatco (approved):

PR #1046, opened 2026-06-25
Fixes #1044 - Fix performance issues parametrization

### LINKED ISSUE #1044: [BUG] Performance issues when using parametrization on a larger report repository
### Library Version

1.1.0

### Python Version

3.11

### Operating System

Linux

### Authentication Method

Azure CLI (az login)

### What is the problem?

Deployments get a lot more slow when having a large respositories, with a bunch of files. In my case I had 58 semantic models and reports, where the reports were using the enhanced metadata format. That means that for a single report we could have over 400 json files.  

Functionally the output stayed correct, but runtime was inflated by redundant local scanning work. This made deploy cycles noticeably heavier in medium-to-large Fabric workspaces and created a scalability bottleneck in day-to-day development and CI.

When deploying the example of 58 objects, It was taking more then 2 hours and wasn't even finished.

### Steps to reproduce

Add objects to your workspace repository with a large number of files. The deployment is almost exponentially growing with each report you add.

### Expected behavior

Deployment performance should scale more predictably with actual work performed. Parameter filters should be evaluated with file-local context where possible, and repository refresh should avoid rediscovering the same filesystem structure repeatedly. The deployment lead time should be almost linear with the number of objects you are deploying.

### Actual behavior

The implementation effectively re-walked significant parts of the repository in paths that run frequently, leading to redundant I/O and path computation. Functional output remained correct, but execution time was inflated due to avoidable repeated traversal.

_No response_

_No response_

10bas10:
## Description

This fix addresses the performance issues when using parametrization on your repository items. The parameter replacement was applied multiple times on the entire repository. 

Another part where some performance issues were applicable is on the retrieving files belonging to items, this has been optimized also for a great part.

This PR fixes this issue:
#1044


10bas10:
@microsoft-github-policy-service agree company="Rabobank"

PR #1053, opened 2026-07-06
Expose cli deploy command in docs

ayeshurun:


docs/how_to/config_deployment.md
```diff
@@ -503,6 +503,52 @@ deploy_with_config(
...
+
+```bash
+fab deploy --config config.yml --target_env test -P config_override='{"core":{"item_types_in_scope":["Notebook"]}}'
+```
+
+<span class="md-h4-nonanchor">Deploy without confirmation prompts (useful in pipelines):</span>
+
+```bash
+fab deploy --config config.yml --target_env dev --force
+```
```
shirasassoon, line 550: What about the bulk publish example?

ayeshurun: will wait until the cli version is published with bulk support

shirasassoon (approved):

PR #1061, opened 2026-07-12
Closes #1059 - Expose supported feature flags method

### LINKED ISSUE #1059: [FEATURE] Expose a function to list all supported feature flags
## Summary

There is currently no public, programmatic way to discover which feature flags `fabric-cicd` supports. Users have to read the docs or source (the `FeatureFlag` enum) to know what values are valid for `append_feature_flag` or the `features` section of a config file.

## Proposal

Add a public `get_supported_feature_flags()` function to the top-level `fabric_cicd` package that returns a sorted list of all supported feature flag string identifiers (the same values accepted by `append_feature_flag` and config files).

## Motivation

- Makes it easy to validate user-provided flags at runtime.
- Enables tooling/UX (e.g. listing available flags, autocompletion, docs generation).
- Provides a single source of truth derived directly from the `FeatureFlag` enum.

## Example

```python
from fabric_cicd import get_supported_feature_flags

get_supported_feature_flags()
# ['continue_on_shortcut_failure', 'disable_workspace_folder_publish', 'enable_bulk_publish', ...]
```

ayeshurun:
## Description

Adds a public get_supported_feature_flags() function to the top-level fabric_cicd package so callers can programmatically discover every supported feature flag instead of reading the source or docs. It returns a sorted list of the flag string identifiers, derived directly from the FeatureFlag enum, so it stays in sync automatically as flags are added or removed. These are the same values accepted by append_feature_flag and the features section of a config file.

The function is exported via __all__, and tests cover completeness against the enum, sorted string output, independence from runtime-appended (unsupported) flags, and round-trip compatibility with append_feature_flag.

```py
from fabric_cicd import get_supported_feature_flags
get_supported_feature_flags()
# ['continue_on_shortcut_failure', 'disable_workspace_folder_publish', ...]
```

## Linked Issue (REQUIRED)

Closes #1059

### DIFF
--- a/src/fabric_cicd/__init__.py
+++ b/src/fabric_cicd/__init__.py
@@ -34,6 +34,25 @@ def append_feature_flag(feature: str) -> None:
     constants.FEATURE_FLAG.add(feature)
 
 
+def get_supported_feature_flags() -> list[str]:
+    """
+    Return all feature flags supported by fabric-cicd.
+
+    The returned values are the string identifiers accepted by
+    ``append_feature_flag`` and the ``features`` section of a config file.
+
+    Returns:
+        A sorted list of supported feature flag string values.
+
+    Examples:
+        Basic usage
+        >>> from fabric_cicd import get_supported_feature_flags
+        >>> get_supported_feature_flags()
+        ['continue_on_shortcut_failure', 'disable_workspace_folder_publish', ...]
+    """
+    return sorted(flag.value for flag in FeatureFlag)
+
+
 def change_log_level(level: str = "DEBUG") -> None:
     """
     Sets the log level for all loggers within the fabric_cicd package. Currently only supports DEBUG.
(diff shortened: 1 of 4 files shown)


src/fabric_cicd/__init__.py
```diff
@@ -34,6 +34,25 @@ def append_feature_flag(feature: str) -> None:
     constants.FEATURE_FLAG.add(feature)
 
 
+def get_supported_feature_flags() -> list[str]:
+    """
+    Return all feature flags supported by fabric-cicd.
```
shirasassoon, line 39: I know this has a docstring, I think for awareness we should document this before or after the feature flag charts on the optional features page

ayeshurun: Great suggestion. Done

shirasassoon (approved):

PR #1077, opened 2026-07-20
Closes #1008 - Add functionality to debug key-value replacement parameters

### LINKED ISSUE #1008: [FEATURE] Validate parameters before deployment
### Use Case / Problem

**Problem**
The current problem I have in my workflow is that for my `key_value_replace` rules I'm unable to get insight into how many files or variables will be touched by a specific rule and if the key I provided is actually valid. 

**Current limitations**
I can only check that after deployment and manually check to see what variables have been touched and what not.

**How this fits into your workflow**
A method that can perform some sort of 'dry run' would be nice. To get an overview which keys in what file are actually going to be changed and what keys were misspecified and thus won't change anything.

### Proposed Solution

**Specific functionality**
A method that can validate a yaml file for a speficied environment.

**Expected behaviour and output**
The method should efficiently check out all json files in the given directory and check what keys can be found and what keys are not found. It should be able to return a simple True or False whether all keys can be found or not. If you specify, it should report all key value pairs if it is found, what the old value is and the replacement value.

**Integration with existing fabric-cicd features**
I can imagine two ways in which this functionality can be integrated. First of all, the function should when validating the parameter file: `_refresh_parameter_file()`, if it is unable to match all keys, it shouldn't fail because the parameters within the file cannot all be validated.
Secondly, for development purposes, the method should be callable by the user, to view the impact of all of the specified `key_value_replace` rules.

### Alternatives Considered

Currently I have my own fabric-cicd wrapper written in which I have implemented such a feature. I have vibe coded it as this is just a simple proof of concept.

```python
def validate_replacements(
        self,
        yaml_file: str = None,
        project_root: str = None,
        environment: str = None,
    ):
        """
        Validate replacement rules efficiently.

        Each JSON file is read only once.
        """

        yaml_file = yaml_file or self.parameter_file_path
        project_root = project_root or self.repository_directory
        environment = environment or self.environment

        with open(yaml_file, "r", encoding="utf-8") as f:
            config = yaml.safe_load(f)

        rules = []

        # Pre-compile JSONPath expressions
        for idx, rule in enumerate(config.get("key_value_replace", [])):
            rules.append(
                {
                    "rule_id": idx,
                    "json_path": rule["find_key"],
                    "compiled_path": parse(rule["find_key"]),
                    "file_path": rule.get("file_path"),
                    "new_value": rule.get("replace_value", {}).get(environment),
                    "matches_found": 0,
                }
            )

        results = []

        all_json_files = list(Path(project_root).rglob("*.json"))

        for json_file in all_json_files:

            try:
                with open(json_file, "r", encoding="utf-8") as f:
                    data = json.load(f)

            except Exception as exc:
                results.append(
                    {
                        "file_path": str(json_file),
                        "found": False,
                        "error": f"Cannot load JSON: {exc}",
                    }
                )
                continue

            for rule in rules:

                # Skip files not targeted by the rule
                if rule["file_path"]:

                    target = Path(rule["file_path"])

                    try:
                        if json_file.resolve() != target.resolve():
                            continue
                    except Exception:
                        if str(json_file) != str(target):
                            continue

                matches = rule["compiled_path"].find(data)

                for match in matches:

                    rule["matches_found"] += 1

                    results.append(
                        {
                            "rule_id": rule["rule_id"],
                            "json_path": rule["json_path"],
                            "file_path": str(json_file),
                            "match_location": str(match.full_path),
                            "current_value": match.value,
                            "new_value": rule["new_value"],
                            "found": True,
                        }
                    )

        # Report rules with no matches
        for rule in rules:

            if rule["matches_found"] == 0:

                results.append(
                    {
                        "rule_id": rule["rule_id"],
                        "json_path": rule["json_path"],
                        "file_path": rule["file_path"],
                        "found": False,
                        "new_value": rule["new_value"],
                        "error": "No matches found",
                    }
                )

        return results 
```

Running this method yields something like the following:
```JSON
[
    {
        "rule_id": 0,
        "json_path": "$.displayName",
        "file_path": "ExampleReport.Report/definition/pages/Page001/page.json",
        "match_location": "displayName",
        "current_value": "Home Page",
        "new_value": "Home Page - TEST",
        "found": True
    },
    {
        "rule_id": 1,
        "json_path": "$.visual.query.queryState.Values.projections.0.nativeQueryRef",
        "file_path": "ExampleReport.Report/definition/pages/Page002/visuals/Visual001/visual.json",
        "found": False,
        "new_value": "Average Processing Time",
        "error": "No matches found"
    },
    {
        "rule_id": 2,
        "json_path": "$.visual.query.queryState.Values.projections.0.displayName",
        "file_path": "ExampleReport.Report/definition/pages/Page002/visuals/Visual001/visual.json",
        "found": False,
        "new_value": "Average Processing Time",
        "error": "No matches found"
    }
]

```

As you can see, it works, however, for full integration it should be refined ofcourse. Just a proof of concept that works for my use case right now.

### Impact Assessment

- [x] This would help me personally
- [ ] This aligns with Microsoft Fabric roadmap items

### Implementation Notes

For now, I've only tested this with Report types. I'm not sure how it holds up with other item types.
It will probably require changes to the core class `Parameter` as that would be the logical location for the method.
In a way it can be combined with #984

ChielChiel:
## Description

Add functionality to debug key-value replacement parameters.
Using the new `FabricWorkspace.validate_parameter_replace()` method you can dry-run the key-value replace parameters and see what files and keys are touched during deployment.

## Linked Issue (REQUIRED)
Linked issue: #1008


ChielChiel:

src/fabric_cicd/_parameter/_parameter.py
```diff
@@ -89,6 +89,210 @@ def __init__(
...
+                this_item_name = meta["metadata"]["displayName"]
+            except Exception:
+                continue
+
+            # Examine every file in the item directory and subdirectories
+            # Walk the entire item directory tree for JSON/YAML files
+            for root, _dirs, files in os.walk(item_root):
+                for file_name in files:
+                    if file_name == ".platform":
+                        continue
```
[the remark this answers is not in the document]
ChielChiel: Intentionally did not implement this change. As can be seen in de sample workspace, Fabric items can be nested inside folders but still need to be found. This supposed 'fix' would break that.

ChielChiel:
Hi @yaronpri, is there anything I can/have to do on my end before this PR can get approved? Happy to help

shirasassoon:
> Hi @yaronpri, is there anything I can/have to do on my end before this PR can get approved? Happy to help

Thanks for the contribution! I will review the PR.

src/fabric_cicd/fabric_workspace.py
```diff
@@ -296,6 +296,37 @@ def _get_workspace_pools(self) -> list[dict]:
...
+        parameter_obj = Parameter(
+            repository_directory=self.repository_directory,
+            item_type_in_scope=self.item_type_in_scope,
+            environment=self.environment,
+            parameter_file_name=constants.PARAMETER_FILE_NAME,
+            parameter_file_path=self.parameter_file_path,
+        )
+
+        validated_results = parameter_obj._validate_key_value_replacements(environment=environment, as_dict=as_dict)
+        if as_dict == False:
```
shirasassoon, line 325: Dead branch — both the `if` and the fall-through `return` return the same value, so this can be simplified to a single `return validated_results`. Also prefer `not as_dict` (or `is False`) over `== False`.

src/fabric_cicd/_parameter/_parameter.py
```diff
@@ -89,6 +89,214 @@ def __init__(
...
+                compiled = None
+            compiled_rules.append({
+                "rule_index": idx,
+                "find_key": find_key,
+                "compiled": compiled,
+                # rule-level optional filters (may be str or list)
+                "rule_item_type": rule.get("item_type"),
+                "rule_item_name": rule.get("item_name"),
+                "rule_file_path": rule.get("file_path"),
+                "new_value": (rule.get("replace_value") or {}).get(environment),
```
shirasassoon, line 160: `_ALL_` environment key isn't resolved here. Deployment runs `process_environment_key()` before reading the value, which maps a lone `_ALL_` key to the target environment. Reading `replace_value` raw means an `_ALL_`-based rule reports `new_value=None` and gets counted under "value not defined for environment", which is a false negative. Consider routing through `process_environment_key()`.

src/fabric_cicd/_parameter/_parameter.py
```diff
@@ -89,6 +89,214 @@ def __init__(
...
+                "matches_found": 0,
+            })
+
+        results: list[dict] = []
+
+        # Walk through all files in the repository_directory, looking for JSON/YAML files to apply the rules to
+        # First pass: find all item root directories (those containing the .platform file)
+        item_roots = []
```
shirasassoon, line 168: This walks every item that has a `.platform` file, ignoring `self.item_type_in_scope`. Real deployment only processes in-scope item types, so the validator can report matches (and `all_matched=True`) for items a deploy will never touch — a false positive. Filter `this_item_type` against `item_type_in_scope` before scanning.

src/fabric_cicd/_parameter/_parameter.py
```diff
@@ -89,6 +89,214 @@ def __init__(
...
+                for file_name in files:
+                    if file_name == ".platform":
+                        continue
+
+                    file_path_abs = Path(root) / file_name
+                    rel_path = str(file_path_abs.relative_to(self.repository_directory))
+
+                    try:
+                        raw = file_path_abs.read_text(encoding="utf-8")
+                    except Exception:
```
shirasassoon, line 195: Silent `except Exception: continue` — a malformed file that deployment would fail on is silently skipped here, lowering fidelity. At minimum log at debug level so users can see what was skipped and why.

src/fabric_cicd/_parameter/_parameter.py
```diff
@@ -89,6 +89,214 @@ def __init__(
...
+        if not as_dict:
+            # Return results as human-readable string e.g., "3/5 rules matched, 2 rules had no matches. For 1 rules, the new value was not defined for environment PPE."
+            return (
+                all_matched,
+                f"{rules_with_matches}/{total_rules} rules matched, {rules_without_matches} rules had no matches. For {rules_without_new_value} rules, the new value was not defined for environment {environment}.",
+            )
+
+        return all_matched, results
+
+    def _rule_filter_matches(
```
shirasassoon, line 264: This reimplements file-path filtering with `fnmatch` on a normalized POSIX string, whereas deployment resolves paths via `process_input_path()` and compares with `Path` equality in `check_replacement`/`_find_match`. `fnmatch` semantics differ (`*` crosses `/`, no anchoring, different backslash handling), so a rule can match/skip differently than an actual deploy. Please reuse the existing `extract_parameter_filters` + `check_replacement` path instead of a parallel implementation.

src/fabric_cicd/_parameter/_parameter.py
```diff
@@ -89,6 +89,214 @@ def __init__(
...
+        if rule_fp is not None:
+            import fnmatch
+
+            file_path_norm = (file_path or "").replace("\\", "/").lstrip("/")
+            patterns = rule_fp if isinstance(rule_fp, list) else [rule_fp]
+            patterns = [p.replace("\\", "/").lstrip("/") for p in patterns]
+            if not any(fnmatch.fnmatch(file_path_norm, pattern) for pattern in patterns):
+                return False
+
+            # if not fnmatch.fnmatch(file_path.replace("\\", "/"), rule_fp.replace("\\", "/")):
```
shirasassoon, line 295: Remove this commented-out dead code before merge.

shirasassoon (commented):
Thanks for adding this — a dry-run for `key_value_replace` is genuinely useful and the test coverage is broad. My main concern is that the validator **reimplements** deployment matching logic instead of reusing it, so it can report results that diverge from what an actual deploy does (which undercuts the point of a validator). Details inline; summary of the higher-priority items:

1. **`item_type_in_scope` is ignored** — the walk visits every `.platform` item, but deployment only processes in-scope item types, so this can report matches (and `all_matched=True`) for items a real deploy never touches.
2. **`_ALL_` env key not resolved** — deployment runs `process_environment_key()` to map a lone `_ALL_` key to the target env; the validator reads `replace_value` raw and will falsely report `new_value=None` / "value not defined".
3. **`file_path` matching diverges** — deployment uses `process_input_path()` + `Path`-equality (`check_replacement`/`_find_match`); the new code uses `fnmatch` on a POSIX string, which has different anchoring/`/`-crossing semantics.
4. **Silent `except Exception: continue`** on read/parse lowers fidelity vs. a real deploy that would surface the error.

Strong recommendation: reuse the existing `extract_parameter_filters` + `check_replacement` + `replace_key_value` utilities (or refactor them to expose a dry-run path) rather than maintaining parallel logic that will drift. Adding a test that exercises the public `FabricWorkspace.validate_parameter_replace()` entrypoint (plus `_ALL_` and out-of-scope cases) would also help.

shirasassoon (changes requested):

shirasassoon:
@ChielChiel Closing this PR due to inactivity and we are planning on pushing changes that overlap with this change set anyway. We can revisit this later on if needed.

PR #1087, opened 2026-07-27
Fixes #1072 -Fix PyYAML corrupting unquoted HH:MM:SS values

### LINKED ISSUE #1072: [BUG]  live_pool schedule time values corrupted during Environment deployment (PyYAML sexagesimal parsing)
### Library Version

1.2.0

### Python Version

3.11.15

### Operating System

Windows

### Authentication Method

Service principal (secret)

### What is the problem?

When deploying an Environment item that has a  live_pool  schedule configured in  Setting/Sparkcompute.yml , the start_time  and  end_time  fields are silently corrupted before being sent to the Fabric API, causing the publish to fail with:

Error Code: UpdateArtifactFailure. Error Message: Succeeded = False, ErrorInfo = 
(Code = SparkSettingsInvalidClpSchedule, Message = 'Invalid custom live pool schedule EndTime: 64800.00:00:00.')

The cause is that  _process_environment_file  in  _items/_environment.py  performs a  yaml.safe_load  →  yaml.dump  round-trip on  Sparkcompute.yml  whenever  instance_pool_id  is present. PyYAML implements YAML 1.1, which parses unquoted  HH:MM:SS  values as base-60 (sexagesimal) integers:

start_time: 08:00:00  →  yaml.safe_load  →  28800 (int)  →  yaml.dump  →  28800
end_time:   18:00:00  →  yaml.safe_load  →  64800 (int)  →  yaml.dump  →  64800

The Fabric API receives the bare integer  64800  and interprets it as a .NET  TimeSpan  of  64800.00:00:00  (64800 days ≈ 177 years), which fails validation. This affects any Environment item with a  live_pool  schedule, regardless of whether any  spark_pool  parameter replacement is actually needed — the round-trip runs unconditionally whenever  instance_pool_id  appears in the file.

### Steps to reproduce

1. Create an Environment item with a  live_pool  schedule in  Setting/Sparkcompute.yml :

instance_pool_id: <some-guid>
live_pool:
  state: Enabled
  schedule_state: Enabled
  schedule:
    recurrence_type: Daily
    start_time: 08:00:00
    end_time: 18:00:00
    time_zone: Pacific Standard Time

2. Deploy the Environment item using fabric-cicd
3. Observe the  SparkSettingsInvalidClpSchedule  error during the Environment publish step

### Expected behavior

The  start_time  and  end_time  values are preserved as time strings ( 08:00:00 ,  18:00:00 ) and accepted by the Fabric API.

### Actual behavior

PyYAML parses  18:00:00  as the integer  64800  (18×3600).  yaml.dump  outputs  64800 . The Fabric API rejects this as  64800.00:00:00  — an invalid .NET TimeSpan.

Replace  yaml.safe_load  /  yaml.dump  with  ruamel.yaml  (YAML 1.2), which does not perform sexagesimal parsing and preserves the original string representation of time values through the round-trip. Alternatively, the  yaml.dump  call could use a custom representer that re-quotes values that were originally strings.

Workaround (until fixed): quote the time values in  Sparkcompute.yml  so PyYAML treats them as strings:

\# Broken — PyYAML YAML 1.1 parses as integer
start_time: 08:00:00
end_time: 18:00:00

\# Fixed — single quotes force string type through round-trip
start_time: '08:00:00'
end_time: '18:00:00'

Note this workaround must be applied manually to the committed file and will be overwritten if the file is re-exported from the Fabric portal, since the portal writes unquoted time values.

shirasassoon:
## Why

Deploying an Environment item whose `Setting/Sparkcompute.yml` contains a `live_pool` schedule fails with `SparkSettingsInvalidClpSchedule` (for example `Invalid custom live pool schedule EndTime: 64800.00:00:00`). The root cause is PyYAML's YAML 1.1 behavior: unquoted `HH:MM:SS` values are parsed as base-60 (sexagesimal) integers, so `end_time: 18:00:00` becomes `64800` and is re-emitted as a bare integer that the Fabric API interprets as a 64800-day .NET TimeSpan.

This corruption occurs on any code path that round-trips YAML through `yaml.safe_load` / `yaml.dump`. An audit of every YAML call site in `src/` found two such deployment paths. Rather than patch each site independently, this PR centralizes all YAML handling so the whole bug class is fixed in one place and cannot silently reappear.

## Centralized handling (`_common/_yaml_safe.py`)

A new shared module is now the single place YAML is loaded and dumped across the package. `load_yaml` / `dump_yaml` use a sexagesimal-safe `SafeYamlLoader` / `SafeYamlDumper` pair that strips only the base-60 branch from the int/float implicit resolvers. Time strings are preserved end to end (loaded as strings, emitted unquoted to match the Fabric portal format) while normal ints, floats, octal, hex, and binary literals still resolve exactly as before. `dump_yaml` defaults to `sort_keys=False` so key order is preserved. `load_yaml` accepts a string, bytes, or an open stream.

Every YAML call site is routed through these helpers:

- **Environment publisher (`_items/_environment.py`)** - `_process_environment_file` no longer re-serializes `Sparkcompute.yml` at all. Since this path only mutates one known key, the YAML is parsed (via `load_yaml`) solely to look up the `spark_pool` mapping, and the resolved GUID is written back with a targeted single-line replacement of `instance_pool_id`. Every other line (comments, formatting, schedule times) is preserved byte-for-byte, and the file is returned unchanged when no mapping matches.
- **`key_value_replace` parameterization (`_parameter/_utils.py`)** - applies replacements to an arbitrary user-supplied JSONPath, so it genuinely needs to parse and re-dump. It now uses `load_yaml` / `dump_yaml`, preserving times and key order. JSON targets were never affected.
- **Parameter file loading (`_parameter/_parameter.py`)** - `_DuplicateKeyLoader` now inherits from `SafeYamlLoader`, so duplicate-key detection and time-safe loading are combined. This means an unquoted `HH:MM:SS` used as a `replace_value` is now preserved correctly (`18:00:00` stays a string) instead of being corrupted to `64800`.
- **`_common/_check_utils.py` and `_common/_config_validator.py`** - YAML validation reads are routed through `load_yaml` for consistency.

Because consistent time-safe loading now fixes the parameter-file value at parse time, the interim "unquoted time" warning added earlier in this PR's history is removed - it would have been redundant and its message (telling users to quote the value) would have been misleading now that the value is handled correctly automatically.

## Guardrail

A test greps `src/` and fails if a raw `yaml.safe_load` / `yaml.dump` (or a bare `yaml.load` without an explicit safe `Loader=`) is reintroduced outside the shared module, so this class of corruption cannot creep back in.

## Tests

- `tests/test_yaml_safe.py`: `load_yaml` preserves unquoted times, still parses normal scalars, accepts a stream; `dump_yaml` emits times unquoted and preserves key order; plus the guardrail test.
- `tests/test_environment_publish.py`: unquoted `live_pool` times survive while `instance_pool_id` is replaced; files unchanged when no replacement applies.
- `tests/test_parameter_utils.py`: `key_value_replace` on YAML preserves schedule times and key order and still resolves normal int/float/bool types.
- `tests/test_parameter.py`: the parameter loader preserves an unquoted time value as a string.

Full suite passes (1083 passed, 11 skipped); ruff format and lint are clean. No new dependency added.

### DIFF
--- a/src/fabric_cicd/_items/_environment.py
+++ b/src/fabric_cicd/_items/_environment.py
@@ -5,6 +5,7 @@
 
 import logging
 import re
+from typing import Optional
 
 import dpath
 import yaml
@@ -33,6 +34,14 @@ def _process_environment_file(
     using the ``spark_pool`` parameter configuration so that the correct pool
     reference is embedded directly in the YAML sent to the Fabric Items API.
 
+    The YAML is parsed only to look up and match the pool mapping; the resolved
+    pool GUID is then written back with a targeted replacement of the
+    ``instance_pool_id`` line, leaving the rest of the file byte-for-byte
+    unchanged. This avoids a ``yaml.safe_load`` / ``yaml.dump`` round-trip, which
+    (under PyYAML's YAML 1.1 rules) would corrupt other values such as unquoted
+    ``live_pool`` schedule ``HH:MM:SS`` times, which are parsed as sexagesimal
+    integers (see issue #1072).
+
     All other files are returned unchanged.
 
     Args:
@@ -52,18 +61,19 @@ def _process_environment_file(
         return contents
 
     yaml_body = yaml.safe_load(contents)
-    if not isinstance(yaml_body, dict):
+    if not isinstance(yaml_body, dict) or "instance_pool_id" not in yaml_body:
         return contents
 
-    if "instance_pool_id" in yaml_body:
-        yaml_body = _replace_instance_pool_id(fabric_workspace_obj, yaml_body, item.name)
+    resolved_id = _resolve_instance_pool_id(fabric_workspace_obj, yaml_body, item.name)
+    if resolved_id is None or resolved_id == yaml_body["instance_pool_id"]:
+        return contents
 
-    return yaml.dump(yaml_body, default_flow_style=False, sort_keys=False)
+    return _replace_instance_pool_id_line(contents, resolved_id)
 
 
-def _replace_instance_pool_id(fabric_workspace_obj: FabricWorkspace, yaml_body: dict, item_name: str) -> dict:
+def _resolve_instance_pool_id(fabric_workspace_obj: FabricWorkspace, yaml_body: dict, item_name: str) -> Optional[str]:
     """
-    Replace ``instance_pool_id`` in parsed Sparkcompute YAML with a resolved pool GUID.
+    Resolve the target ``instance_pool_id`` GUID for Sparkcompute YAML.
 
     This function reads ``spark_pool`` parameter mappings from
     ``fabric_workspace_obj.environment_parameter`` and finds the entry whose
@@ -72,8 +82,7 @@ def _replace_instance_pool_id(fabric_workspace_obj: FabricWorkspace, yaml_body: 
     otherwise, the mapping applies globally.
 
     The mapped target pool ``name`` and ``type`` are then resolved against the
-    workspace custom pool list returned by the Fabric API, and the resolved pool
-    ``id`` is written back to ``yaml_body["instance_pool_id"]``.
+    workspace custom pool list returned by the Fabric API.
 
     Args:
         fabric_workspace_obj: Workspace context containing environment, parameters,
@@ -82,7 +91,7 @@ def _replace_instance_pool_id(fabric_workspace_obj: FabricWorkspace, yaml_body: 
         item_name: Environment item name used for optional per-item mapping filters.
 
     Returns:
-        The YAML dictionary, updated if a matching mapping is found; otherwise unchanged.
+        The resolved pool GUID if a matching mapping is found; otherwise ``None``.
     """
     from fabric_cicd._parameter._utils import _find_match, process_environment_key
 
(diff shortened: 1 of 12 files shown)


src/fabric_cicd/_items/_environment.py
```diff
@@ -96,15 +99,35 @@ def _replace_instance_pool_id(fabric_workspace_obj: FabricWorkspace, yaml_body:
...
+
+    Returns:
+        The contents with the ``instance_pool_id`` value replaced.
+    """
+    return re.sub(
+        r"^(?P<prefix>instance_pool_id[ \t]*:[ \t]*).*$",
+        lambda m: f"{m.group('prefix')}{resolved_id}",
+        contents,
+        count=1,
+        flags=re.MULTILINE,
```
ayeshurun: is this necessary? can it be a multiline?

shirasassoon: fixed

ayeshurun (approved):

shirasassoon:
abandoning this approach for now.

PR #1094, opened 2026-08-04
Add easy parameterization of Semantic Model Parameters

### LINKED ISSUE #552: Support replacing parameters in Semantic Models by name
### What is the feature?

First of all: thank you for the great work on Fabric CI/CD, it’s already a huge step forward to kick-off the CICD efforts on Fabric!

I have a use case around Semantic Models where I would like to change parameters during deployment.
Currently, the tooling works when you know the value you want to replace, but in practice I often do not know the existing value in advance. Instead, I’d like to target replacements by parameter name. Bonus, same concept but then on connection name.

### Additional context

**Example scenario**
- A Semantic Model contains a parameter DatabaseServer. In different environments, I want to override this with environment-specific values (e.g., sql-dev.contoso.net, sql-prod.contoso.net).
- Today I need to know the old value in order to replace it. In multi-developer or changing environments this is not always reliable.

**Proposed solution**
Allow the key-value replacement function to support:
- Lookups by parameter name (not just by existing value).
- BONUS: Lookups by connection name when updating external connections.

**YAML-style example** (aligned with current key_value_replace syntax)
Image

**Benefits**
- More reliable environment deployments.
- Reduced dependency on knowing exact values upfront.
- Easier multi-environment support (dev/test/prod).

**References**
Reddit discussion that led me here: https://www.reddit.com/r/MicrosoftFabric/comments/1n9d9t4/comment/nee89mv/?context=1

gbrueckl:
## Description

This PR adds a new parameterization option specifically for Semantic models:
```yaml
semantic_model_parameter:
  - semantic_model_name: "MyModel"
    parameter_name: "myParameter"
    parameter_value:
      TEST: "mytestserver"
      PROD: "prodserver
```

its basically a wrapper for `find_replace` and is internally also converted to a new entry in `find_replace`.

## Linked Issue

Fixes #552


gbrueckl:

PR #1109, opened 2026-08-16
Resolves #1105 Fix cluster uri replacement in KQL Queryset item

### LINKED ISSUE #1105: [BUG] KQL Querysets deployment error when clusterUri is missing
### What is the question?

Is clusterUri value required or not? The documentation mentions "fabric-cicd handles this automatically" when the queryset is attached to a database within the same workspace.

Image

However, when I tried deploying, I get the following error message.

> Failed to publish KQLQueryset trigger__sn_edp: Cannot find the KQL Database source with name None as it is not yet deployed.‌

Alternatively, if it is required and is actually not automatically handled, is there a way to avoid the workspace-git integration from blanking it out/ changing the details when it detects the db is within the same workspace?

### Context

I have identified that the failed deployment issue has to do with the missing clusterUri value by attempting to deploy 2 querysets, both of which reference the same database within the same deployment-target workspace.
- trigger__cpa_dw: with clusterUri and databaseItemId = actual db item ID - a change I manually introduced; deploys successfully
- trigger__sn_edp: empty clusterUri and databaseItemId = logical ID - change introduced by workspace-git integration seemingly upon detecting the referenced KQL DB is within the same workspace; this fails with the error message mentioned above.

Image

Image

If this was intended behavior, would suggest updating the error message as it does not mention what it's actually missing/expecting, i.e., clusterUri value.

Have tried different versions of fabric-cicd from 1.0.0, 1.1.0, 1.2.0 and the current one 1.3.0.

### Library Version

_No response_

### Operating System

_No response_

### Have you checked the documentation?

- [x] I have searched existing GitHub issues
- [x] I have reviewed the fabric-cicd documentation

### Additional Information

_No response_

shirasassoon:
This pull request updates the logic for replacing empty cluster URIs in KQL Queryset files, improving the accuracy and robustness of how KQL Database items are matched and referenced. The changes ensure that the system uses repository items and logical IDs instead of deployed items and names, resulting in more reliable behavior during parsing and error handling.

**Improvements to KQL Database item lookup and error handling:**

* The function now retrieves KQL Database items from `fabric_workspace_obj.repository_items` instead of `deployed_items`, ensuring it works with repository state rather than deployment state.
* When resolving empty cluster URIs, the code matches KQL Database items by their logical ID (`databaseItemId`) instead of by name, making the lookup more robust and less error-prone.
* Improved error messages and log statements to reference the database's logical ID and name, providing clearer diagnostics when a database or its cluster URI cannot be found. [[1]](diffhunk://#diff-996cee517cea88bd9e7c873c7bb485859c629fd192a348d4ac41456264669864L53-R69) [[2]](diffhunk://#diff-996cee517cea88bd9e7c873c7bb485859c629fd192a348d4ac41456264669864L79-R89)

### DIFF
--- a/src/fabric_cicd/_items/_kqlqueryset.py
+++ b/src/fabric_cicd/_items/_kqlqueryset.py
@@ -57,19 +57,24 @@ def replace_cluster_uri(fabric_workspace_obj: FabricWorkspace, file_obj: File) -
     for data_source in data_sources:
         if data_source.get("clusterUri") == "":
             # Get the logical ID of the source KQL database
-            database_id = data_source.get("databaseItemId")
-            logger.debug(f"Found empty cluster URI for database with ID '{database_id}'")
+            database_logical_id = data_source.get("databaseItemId")
+            logger.debug(f"Found empty cluster URI for database with ID '{database_logical_id}'")
 
             database_item = next(
-                (item for item in database_items.values() if item.logical_id == database_id),
+                (item for item in database_items.values() if item.logical_id == database_logical_id),
                 None,
             )
-
-            if not database_item or not database_item.guid:
-                msg = f"Cannot find a deployed KQL Database with logical ID '{database_id}'."
+            if not database_item:
+                msg = f"Cannot find a KQL Database source with logical ID '{database_logical_id}' in the repository."
                 raise ParsingError(msg, logger)
 
+            database_item_name = database_item.name
             database_item_guid = database_item.guid
+
+            if not database_item_guid:
+                msg = f"Cannot find the KQL Database source with name '{database_item_name}' as it is not yet deployed."
+                raise ParsingError(msg, logger)
+
             # Get the cluster URI of the KQL database
             kqldatabase_data = fabric_workspace_obj.endpoint.invoke(
                 method="GET",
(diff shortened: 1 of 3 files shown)


aviatco (approved):

src/fabric_cicd/_items/_kqlqueryset.py
```diff
@@ -50,21 +50,31 @@ def replace_cluster_uri(fabric_workspace_obj: FabricWorkspace, file_obj: File) -
...
+    # Get the KQL Database items from the repository items
+    database_items = fabric_workspace_obj.repository_items.get(ItemType.KQL_DATABASE.value, {})
 
     # If the cluster URI is empty, replace it with the cluster URI of the KQL database
     for data_source in data_sources:
         if data_source.get("clusterUri") == "":
-            database_item_name = data_source.get("databaseItemName")
-            logger.debug(f"Found empty cluster URI for database '{database_item_name}'")
+            # Get the logical ID of the source KQL database
+            database_logical_id = data_source.get("databaseItemId")
```
aviatco, line 60: does the logical_id always exists?

ayeshurun (approved):

src/fabric_cicd/_items/_kqlqueryset.py
```diff
@@ -50,21 +50,31 @@ def replace_cluster_uri(fabric_workspace_obj: FabricWorkspace, file_obj: File) -
         logger.debug("No data sources found in KQL Queryset.")
         return file_obj.contents
 
-    # Get the KQL Database items from the deployed items
-    database_items = fabric_workspace_obj.deployed_items.get(ItemType.KQL_DATABASE.value, {})
+    # Get the KQL Database items from the repository items
+    database_items = fabric_workspace_obj.repository_items.get(ItemType.KQL_DATABASE.value, {})
```
ayeshurun, line 54: is the bug becuase we used deployed_items instead of repository_tems?

shirasassoon: The bug is that we were using the database_name property from the source control file to obtain the cluster uri, but there was a change in the source control file and that property is no longer present. We needed to switch to using the databaseitemid property which is the logical id of the database and therefore we need to use the repository items dict instead of the deployed items to retrieve the cluster uri via logical id.

src/fabric_cicd/_items/_kqlqueryset.py
shirasassoon: It should, this is a defensive check.

PR #1116, opened 2026-08-31
Closes #1117 Support dynamic replacement variables in bulk publish mode

### LINKED ISSUE #1117: [FEATURE] Support Dynamic Replacement Variables in Bulk Publish Mode
### Use Case / Problem

Support all dynamic replacement variables to be used in bulk publishing. 

### Proposed Solution

Use batched approach based on variable dependencies. 

### Alternatives Considered

_No response_

### Impact Assessment

- [x] This would help me personally
- [x] This aligns with Microsoft Fabric roadmap items

### Implementation Notes

_No response_

shirasassoon:
## Summary

Enables native dynamic replacement variable (`$workspace.*` / `$items.*`) support in the **bulk publish** path. Previously, bulk publish fell back to serial deployment whenever the parameter file contained any dynamic variables. Now bulk publish resolves them natively by ordering deployment into dependency tiers.

## Approach

- **New `_bulk_publish_dependencies` module** — builds a dependency graph from same-workspace `$items.*` references, topologically sorts items into tiers via Kahn's algorithm (raising on cycles), and detects unfiltered item references.
- **Tiered execution in `publish_all_bulk`** — instead of a single bulk API call, executes one bulk call per tier, calling `_refresh_deployed_items()` and trimming the resolution cache between tiers so each tier resolves against freshly deployed state.
- **Bulk-scoped resolution cache** (`_dynamic_var_cache`) wired into `extract_replace_value` to avoid redundant resolution within a bulk run; `$workspace.*` keys are retained across tiers while `$items.*` keys are dropped after each tier.
- **Narrowed serial fallback gate** — only an *unfiltered* current-workspace `$items.*` replace_value forces serial fallback (its dependency scope can't be narrowed without a filter). Cross-workspace item vars, `$workspace.*` vars, and dynamic `find_value`s now stay on the bulk path.
- Reuses the canonical `parse_dynamic_variable` parser.

## Asynchronous SQL endpoint / query URI handling

SQL endpoints (`sqlendpoint` / `sqlendpointid`) and the Eventhouse query URI (`queryserviceuri`) are provisioned **asynchronously** after item creation. In serial publishing this is awaited by `check_sqlendpoint_provision_status`, but the bulk path never calls `publish_one`, so a downstream tier could otherwise resolve these to empty strings.

- New `FabricWorkspace._wait_for_item_attribute_provisioning` — a generalized poll (Lakehouse / Mirrored Database / Warehouse / SQL Database / Eventhouse) that waits until the async attribute is available, with terminal-failure detection where the API exposes a provisioning status.
- `get_async_provisioned_dependencies` maps to-be-published source items to the async attributes referenced on them.
- `publish_all_bulk` waits for those attributes on just-deployed source items **before** the between-tier refresh, so the next tier resolves real values.

## Docs

Updated the optional features page to document dynamic replacement variable support in bulk publish, the tiered publishing model, async SQL endpoint handling, and the single remaining fallback condition (unfiltered current-workspace `$items.*` replace values).

## Testing

- `tests/test_bulk_publish.py` — reference parsing, dependency graph, batch computation, tiered execution, async-provisioned dependency mapping, provisioning-wait polling/failure, and the tiered wait hook
- `tests/test_parameter.py` + `tests/test_parameter_utils.py` — confirm cache hooks don't regress non-bulk paths
- Full suite green; `ruff check` / `ruff format` clean

---

> **Note:** Linked issue and changelog entry to follow.

### DIFF
--- a/src/fabric_cicd/_items/_base_publisher.py
+++ b/src/fabric_cicd/_items/_base_publisher.py
@@ -306,17 +306,25 @@ class ItemPublisher(Publisher):
     @staticmethod
     def publish_all_bulk(fabric_workspace_obj: "FabricWorkspace") -> list["ItemPublisher"]:
         """
-        Execute a single bulk publish across all item types in scope.
+        Execute bulk publish across all item types in scope, with batched (tiered) execution
+        when dynamic replacement variable dependencies require it.
 
         Lifecycle:
             1. pre_publish_all() — per type
             2. get_items_to_publish() — per type, collect into one pool
-            3. _publish_items() — single cross-type API call
-            4. post_publish_all() — per type
+            3. Build dependency graph from $items.* dynamic replacement variables
+            4. Compute publish batches (topological sort)
+            5. For each batch: _publish_items() → refresh deployed items between batches
+            6. post_publish_all() — per type
 
         Returns:
             List of publishers that have async checks pending.
         """
+        from fabric_cicd._items._bulk_publish_dependencies import (
+            build_dynamic_variable_dependency_graph,
+            compute_publish_batches,
+        )
+
         publishers: list[ItemPublisher] = []
         items_with_context: list[tuple[str, Item, ItemPublisher]] = []
         skipped_items: list[str] = []
@@ -348,19 +356,45 @@ class ItemPublisher(Publisher):
             items_with_context.extend(publishable_items)
             publishers.append(publisher)
 
-        # Phase 2: Single bulk API call (publisher context used inside for per-item transforms)
+        # Phase 2: Batched bulk API calls (publisher context used inside for per-item transforms)
         if items_with_context:
             if skipped_items:
                 logger.info(f"Skipping {len(skipped_items)} item(s) due to publish filters")
 
-            item_count = len(items_with_context)
-            if item_count > constants.BULK_ITEM_COUNT_LIMIT:
-                msg = (
-                    f"Bulk publish item count ({item_count}) exceeds the API limit "
-                    f"of {constants.BULK_ITEM_COUNT_LIMIT} items."
+            # Build a dependency graph from dynamic replacement variables when the parameter file has them.
+            # An empty graph (the common case) yields a single batch, identical to a single bulk call.
+            dependency_edges: list[tuple[str, str]] = []
+            if fabric_workspace_obj.contains_param_vars:
+                publish_item_keys = {f"{item.type}.{item_name}" for item_name, item, _publisher in items_with_context}
+                dependency_edges = build_dynamic_variable_dependency_graph(fabric_workspace_obj, publish_item_keys)
+
+            batches = compute_publish_batches(items_with_context, dependency_edges)
+
+            for batch_index, batch_items in enumerate(batches):
+                batch_count = len(batch_items)
+                if batch_count > constants.BULK_ITEM_COUNT_LIMIT:
+                    msg = (
+                        f"Bulk publish batch {batch_index} item count ({batch_count}) exceeds the API limit "
+                        f"of {constants.BULK_ITEM_COUNT_LIMIT} items."
+                    )
+                    raise InputError(msg, logger)
+
+                if len(batches) > 1:
+                    batch_names = sorted(f"{item.type}: {name}" for name, item, _ in batch_items)
+                    logger.info(f"Publishing batch {batch_index + 1}/{len(batches)}: {batch_names}")
+
+                fabric_workspace_obj._publish_items(
+                    batch_items, skipped_items=skipped_items if batch_index == 0 else None
                 )
-                raise InputError(msg, logger)
-            fabric_workspace_obj._publish_items(items_with_context, skipped_items=skipped_items)
+
+                # Between batches: refresh deployed items so the next batch can resolve $items.* variables
+                # to real GUIDs. Drop resolved $items.* cache entries (deployed state changed) while keeping
+                # $workspace.* entries, which reference other workspaces that do not change here.
+                if batch_index < len(batches) - 1:
+                    fabric_workspace_obj._dynamic_var_cache = {
+                        k: v for k, v in fabric_workspace_obj._dynamic_var_cache.items() if k.startswith("$workspace.")
+                    }
+                    fabric_workspace_obj._refresh_deployed_items()
 
         # Phase 3: Post-hooks per type
         for publisher in publishers:
--- a/src/fabric_cicd/_items/_base_publisher.py
+++ b/src/fabric_cicd/_items/_base_publisher.py
@@ -372,6 +372,8 @@ class ItemPublisher(Publisher):
 
             batches = compute_publish_batches(items_with_context, dependency_edges)
 
+            # Validate every batch against the API item-count limit before publishing any of them,
+            # so an oversized batch fails fast instead of leaving a partial (batch-by-batch) deployment.
             for batch_index, batch_items in enumerate(batches):
                 batch_count = len(batch_items)
                 if batch_count > constants.BULK_ITEM_COUNT_LIMIT:
@@ -381,6 +383,7 @@ class ItemPublisher(Publisher):
                     )
                     raise InputError(msg, logger)
 
+            for batch_index, batch_items in enumerate(batches):
                 if len(batches) > 1:
                     logger.info(f"Publishing batch {batch_index + 1}/{len(batches)}")
                     logger.debug(
@@ -396,14 +399,15 @@ class ItemPublisher(Publisher):
                 # Refresh deployed items and invalidate cached $items.* values between batches
                 if batch_index < len(batches) - 1:
                     # Wait for async attributes needed by later batches before refreshing
-                    for b_item_name, b_item, _b_publisher in batch_items:
-                        referenced_attrs = async_source_map.get(f"{b_item.type}.{b_item_name}")
-                        if not referenced_attrs:
-                            continue
-                        for attribute_name in sorted(referenced_attrs):
-                            fabric_workspace_obj._wait_for_item_attribute_provisioning(
-                                b_item.type, b_item.guid, b_item_name, attribute_name
-                            )
+                    provisioning_tasks = (
+                        (item, item_name, attribute_name)
+                        for item_name, item, _publisher in batch_items
+                        for attribute_name in sorted(async_source_map.get(f"{item.type}.{item_name}", ()))
+                    )
+                    for item, item_name, attribute_name in provisioning_tasks:
+                        fabric_workspace_obj._wait_for_item_attribute_provisioning(
+                            item.type, item.guid, item_name, attribute_name
+                        )
 
                     fabric_workspace_obj._dynamic_var_cache = {
                         k: v for k, v in fabric_workspace_obj._dynamic_var_cache.items() if k.startswith("$workspace.")
--- a/src/fabric_cicd/_items/_base_publisher.py
+++ b/src/fabric_cicd/_items/_base_publisher.py
@@ -372,13 +372,12 @@ class ItemPublisher(Publisher):
 
             batches = compute_publish_batches(items_with_context, dependency_edges)
 
-            # Validate every batch against the API item-count limit before publishing any of them,
-            # so an oversized batch fails fast instead of leaving a partial (batch-by-batch) deployment.
+            # Validate all batch sizes before publishing to prevent partial deployment
             for batch_index, batch_items in enumerate(batches):
                 batch_count = len(batch_items)
                 if batch_count > constants.BULK_ITEM_COUNT_LIMIT:
                     msg = (
-                        f"Bulk publish batch {batch_index} item count ({batch_count}) exceeds the API limit "
+                        f"Bulk publish batch {batch_index + 1} item count ({batch_count}) exceeds the API limit "
                         f"of {constants.BULK_ITEM_COUNT_LIMIT} items."
                     )
                     raise InputError(msg, logger)
--- a/src/fabric_cicd/_items/_base_publisher.py
+++ b/src/fabric_cicd/_items/_base_publisher.py
@@ -323,6 +323,7 @@ class ItemPublisher(Publisher):
         from fabric_cicd._items._bulk_publish_dependencies import (
             build_dynamic_variable_dependency_graph,
             compute_publish_batches,
+            get_async_provisioned_dependencies,
         )
 
         publishers: list[ItemPublisher] = []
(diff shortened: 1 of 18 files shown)


src/fabric_cicd/_items/_base_publisher.py
```diff
@@ -348,19 +357,58 @@ def publish_all_bulk(fabric_workspace_obj: "FabricWorkspace") -> list["ItemPubli
...
+            batches = compute_publish_batches(items_with_context, dependency_edges)
+
+            for batch_index, batch_items in enumerate(batches):
+                batch_count = len(batch_items)
+                if batch_count > constants.BULK_ITEM_COUNT_LIMIT:
+                    msg = (
+                        f"Bulk publish batch {batch_index} item count ({batch_count}) exceeds the API limit "
+                        f"of {constants.BULK_ITEM_COUNT_LIMIT} items."
+                    )
+                    raise InputError(msg, logger)
```
ayeshurun, line 383: i believe this check and error should be raised when we build the batches ? inside the compute_publish_bathces.

src/fabric_cicd/_items/_base_publisher.py
```diff
@@ -348,19 +357,58 @@ def publish_all_bulk(fabric_workspace_obj: "FabricWorkspace") -> list["ItemPubli
...
+                if batch_index < len(batches) - 1:
+                    # Wait for async attributes needed by later batches before refreshing
+                    for b_item_name, b_item, _b_publisher in batch_items:
+                        referenced_attrs = async_source_map.get(f"{b_item.type}.{b_item_name}")
+                        if not referenced_attrs:
+                            continue
+                        for attribute_name in sorted(referenced_attrs):
+                            fabric_workspace_obj._wait_for_item_attribute_provisioning(
+                                b_item.type, b_item.guid, b_item_name, attribute_name
+                            )
```
ayeshurun:
```suggestion
provisioning_tasks = (
    (item, item_name, attribute_name)
    for item_name, item, _publisher in batch_items
    for attribute_name in sorted(
        async_source_map.get(f"{item.type}.{item_name}", ())
    )
)

for item, item_name, attribute_name in provisioning_tasks:
    fabric_workspace_obj._wait_for_item_attribute_provisioning(
        item.type,
        item.guid,
        item_name,
        attribute_name,
    )

```

src/fabric_cicd/_items/_base_publisher.py
ayeshurun, line 1: do we want to update the item type agent accordingly? (could be done in a separate PR)

src/fabric_cicd/_items/_base_publisher.py
```diff
@@ -348,19 +358,65 @@ def publish_all_bulk(fabric_workspace_obj: "FabricWorkspace") -> list["ItemPubli
...
+                if len(batches) > 1:
+                    logger.info(f"Publishing batch {batch_index + 1}/{len(batches)}")
+                    logger.debug(
+                        "Publishing batch %s items: %s",
+                        batch_index + 1,
+                        sorted(f"{item.type}: {name}" for name, item, _ in batch_items),
+                    )
+
+                fabric_workspace_obj._publish_items(
+                    batch_items, skipped_items=skipped_items if batch_index == 0 else None
```
ayeshurun, line 392: why only for batch_index ==0?

src/fabric_cicd/_items/_bulk_publish_dependencies.py
```diff
@@ -0,0 +1,359 @@
...
+        batches.append(batch)
+        current_reps = sorted(next_reps, key=lambda rep: key_index[rep])
+
+    # Unprocessed groups belong to a dependency cycle
+    if len(processed) < len(groups):
+        cycle_keys = sorted(member for rep in (set(groups) - processed) for member in groups[rep])
+        msg = f"Circular dynamic variable dependency detected among: {', '.join(cycle_keys)}"
+        raise InputError(msg, logger)
+
+    return batches
```
ayeshurun: here we can throw an error for batches size limit instead of in the parent method who calls it.

src/fabric_cicd/_items/_bulk_publish_dependencies.py
```diff
@@ -214,30 +217,27 @@ def build_logical_reference_edges(
...
+                continue
 
-        for logical_id, referenced_key in logical_id_to_key.items():
-            if referenced_key != key and logical_id in content:
-                edge = (key, referenced_key) if key < referenced_key else (referenced_key, key)
-                edges.setdefault(edge, None)
+            for match in _GUID_REFERENCE_PATTERN.finditer(content):
+                referenced_key = logical_id_to_key.get(match.group(0).lower())
+                if referenced_key is not None and referenced_key != key:
+                    edge = (key, referenced_key) if key < referenced_key else (referenced_key, key)
```
ayeshurun: does it makes sense to fall into the 'else' condition? i mean if i'm inside the 3rd inner for lopo, it should always end with (referencing_key, referneced_key), isn't it?

src/fabric_cicd/_items/_bulk_publish_dependencies.py
```diff
@@ -0,0 +1,359 @@
...
+        for file in getattr(item, "item_files", []):
+            content = getattr(file, "contents", None)
+            if getattr(file, "type", None) != "text" or not isinstance(content, str):
+                continue
+
+            for match in _GUID_REFERENCE_PATTERN.finditer(content):
+                referenced_key = logical_id_to_key.get(match.group(0).lower())
+                if referenced_key is not None and referenced_key != key:
+                    edge = (key, referenced_key) if key < referenced_key else (referenced_key, key)
+                    edges.add(edge)
```
ayeshurun: 3 for loops is expensive. can we try to optimize it?

src/fabric_cicd/_items/_base_publisher.py
```diff
@@ -306,25 +302,25 @@ def get_orphaned_items(
...
+            5. For each batch: _publish_items() → refresh deployed items between stages
             6. post_publish_all() — per type
 
         Returns:
             List of publishers that have async checks pending.
         """
         from fabric_cicd._items._bulk_publish_dependencies import (
             build_dynamic_variable_dependency_graph,
-            build_logical_reference_edges,
+            build_logical_id_links,
```
ayeshurun, line 325: much more clearer :)

src/fabric_cicd/_items/_bulk_publish_dependencies.py
```diff
@@ -1,359 +1,630 @@
...
-            continue
-
-        # Record only source items created in this publish operation
-        key = f"{parsed.item_type}.{parsed.item_name}"
-        if key in publish_item_keys:
-            result.setdefault(key, set()).add(parsed.attribute)
-
-    return result
+    items_by_representative: dict[str, list[str]]  # {representative_item_key: [item_key, ...]}
+    representative_by_item: dict[str, str]  # {item_key: representative_item_key}
```
ayeshurun, line 56: consider maybe

representative_by_item_key

src/fabric_cicd/_items/_bulk_publish_dependencies.py
```diff
@@ -1,359 +1,630 @@
...
+                    link = (
+                        (referencing_key, referenced_key)
+                        # Canonicalize the pair for deduplication
+                        if referencing_key < referenced_key
+                        else (referenced_key, referencing_key)
+                    )
+                    links.add(link)
+
+    # Return the list of unique logical-ID-linked item pairs
+    # Example: [("DataPipeline.Sample_PL", "Notebook.Sample_NB"), ("Notebook.Sample_NB", "Environment.Sample_ENV")]
```
ayeshurun, line 231: isn't the format like `<item name>.<item type>`?

```py
Example: [("Sample_PL.DataPipeline", "Sample_NB.Notebook"), ("Sample_NB.Notebook", "Sample_ENV.Environment")]
```

ayeshurun, line 231: btw what will be in case of two items with the same name? how do we determine the name? by the platform file? or directory name?

src/fabric_cicd/_items/_bulk_publish_dependencies.py
```diff
@@ -1,359 +1,630 @@
...
-        if node_a in publish_item_keys and node_b in publish_item_keys:
-            union(node_a, node_b)
+    logger.debug("Bulk publish items in input order: %s", ordered_keys)
+
+    item_key_index = {key: index for index, key in enumerate(ordered_keys)}
+
+    # Build publish units based on logical ID links
+    publish_units = _build_publish_units(ordered_keys, logical_id_links, item_key_index)
+    # Build publish stages based on dependency edges
+    publish_stages = _build_publish_stages(publish_units, dependency_edges, item_key_index)
```
ayeshurun, line 280: what does it mean `stage` in `publish_stages`?

src/fabric_cicd/_items/_bulk_publish_dependencies.py
```diff
@@ -1,359 +1,630 @@
...
-            logical_id_to_key[logical_id] = f"{item.type}.{item_name}"
+        logical_id_to_item_key[item.logical_id.lower()] = f"{item.type}.{item_name}"
 
     # A reference requires at least two distinct logical IDs to relate
-    if len(logical_id_to_key) < 2:
+    if len(logical_id_to_item_key) < 2:
         return []
 
-    edges: dict[tuple[str, str], None] = {}
+    # Identify logical ID bindings between items based on their file contents (skip .platform files)
```
ayeshurun, line 209: can you please add in this comment an example of what referencing_key and referenced_key are

src/fabric_cicd/_items/_bulk_publish_dependencies.py
```diff
@@ -1,359 +1,630 @@
...
 
-    # Collect items with the same root key into one publish group, preserving input order
-    groups: dict[str, list[str]] = {}
+    # Form a publish batch from each dependency stage
+    batches: list[list[tuple[str, object, object]]] = []
+    for stage_index, stage in enumerate(publish_stages):
+        batch: list[tuple[str, object, object]] = []
+        # Extract all items from the unit associated with each representative in the given stage
+        for representative in stage:
+            for item_key in publish_units.items_by_representative[representative]:
```
ayeshurun, line 288: are we sure `representative` is a valid key in `items_by_representative`? consider

```py
if representative in publish_units.items_by_representative:
   for item_key in publish_units.items_by_representative[representative]:
```

src/fabric_cicd/_items/_base_publisher.py
shirasassoon: To prevent redundant logging, we only want to print the list of skipped items once.

src/fabric_cicd/_items/_bulk_publish_dependencies.py
shirasassoon: Updated

src/fabric_cicd/_items/_bulk_publish_dependencies.py
shirasassoon: This format is used consistently in this file when creating both `logical_id_to_item_key` and `referencing_key`. The tests also expect values such as `("DataPipeline.C", "Lakehouse.B")`.

shirasassoon: The format here is intentionally `<item type>.<item name>`, rather than the repository directory convention of `<item name>.<item type>`. The item name comes from `.platform`’s `metadata.displayName`, not from the directory name. Items with the same name but different types remain distinct, for example, `Report.Sales` and `SemanticModel.Sales`. Although two items with the same type and display name would resolve to the same dictionary and graph key, Fabric does not allow creating two items of the same type with the same name, so this collision is not a concern for a valid workspace.

src/fabric_cicd/_items/_bulk_publish_dependencies.py
shirasassoon: `publish_stages` is a list of dependency layers. Each stage is a list of **publish-unit representative keys** that can be published together because all their dependencies were satisfied by earlier stages.

For example:
```
publish_stages = [
    ["Lakehouse.Source"],  # Stage 0
    ["Environment.Python"],  # Stage 1
]
```
In:
`for stage_index, stage in enumerate(publish_stages):`

* stage_index is the zero-based position: 0, 1.
* stage is the current list, such as ["Lakehouse.Source"].

The stages must be processed sequentially:

1. Publish the lakehouse.
2. Publish the environment (includes environment and notebook and pipeline) that depends on it.

A representative can stand for multiple items when logical-ID links force those items into the same publish unit. The inner loops expand each representative into its actual items and create one API batch per stage in `compute_publish_batches`.

src/fabric_cicd/_items/_bulk_publish_dependencies.py
shirasassoon: Yes, representative is guaranteed to be a key in `items_by_representative`:

1. `_build_publish_stages` initializes `in_degree` and dependents exclusively from `publish_units.items_by_representative`.
2. `current_units`, which becomes each stage, is selected from those keys.
3. Later stages only receive values from dependents, which is also built from those same publish-unit representatives.

Adding:
`if representative in publish_units.items_by_representative:`

would be undesirable because an invariant violation would silently omit items from the deployment batch. Direct indexing is safer: an unexpected inconsistency raises `KeyError` instead of producing a seemingly successful but incomplete deployment.

src/fabric_cicd/_items/_bulk_publish_dependencies.py
ayeshurun: so 'stage' can be referred as 'batch'?

ayeshurun (approved):

shirasassoon:
@copilot resolve the merge conflicts in this pull request

src/fabric_cicd/_items/_bulk_publish_dependencies.py
shirasassoon: Yes, you can think of it as an ordered batch or staged batch.

src/fabric_cicd/_items/_bulk_publish_dependencies.py
shirasassoon: Also this order mirrors the dynamic variable order format.

src/fabric_cicd/_items/_bulk_publish_dependencies.py
shirasassoon: Clarified

PR #1131, opened 2026-09-15
Fixes #1129 - Use plain OS variable name for $ENV: environment variable replacement

### LINKED ISSUE #1129: [QUESTION] Environment variable replacement requires `$ENV:` in the actual OS variable name
### What is the question?

The `enable_environment_variable_replacement` feature flag is meant to let users reference environment variables in `parameter.yml` using a `$ENV:VAR_NAME` token.

However, the current implementation appears to require the actual OS environment variable to be named with the `$ENV:` prefix. This seems impractical: AFAIK, no shell or CI system will allow a variable name that starts with `$ENV:`. So, it's effectively unusable outside of manually poking `os.environ` in a Python wrapper.

Would it be better to keep the plain OS variable name (`VAR_NAME`) and use `$ENV:` only as the in-file token prefix? Or am I missing something?

### Context

We currently run fabric-cicd via the fabric-cli command `fab deploy`, so we can only use real OS environment variables—we cannot set keys on `os.environ` from a Python wrapper.

### Library Version

_No response_

### Operating System

Not applicable

### Have you checked the documentation?

- [x] I have searched existing GitHub issues
- [x] I have reviewed the fabric-cicd documentation

### Additional Information

_No response_

shirasassoon:
## Summary

Fixes #1129.

The `enable_environment_variable_replacement` feature flag was effectively unusable. The resolver in `replace_variables_in_parameter_file` only replaced an in-file `$ENV:VAR_NAME` token when the *actual OS environment variable* was literally named `$ENV:VAR_NAME`:

```python
env_vars = {k[len("$ENV:") :]: v for k, v in os.environ.items() if k.startswith("$ENV:")}
```

No shell or CI system (nor `fab deploy`) can set an environment variable whose name begins with `$ENV:`, so the replacement never fired in practice.

## Fix

`$ENV:` is now treated as the **in-file token prefix only**. The resolver scans the parameter file for `$ENV:<name>` tokens and looks up the **plain** OS environment variable name:

```python
for var_name in set(re.findall(r"\$ENV:(\w+)", raw_file)):
    if var_name in os.environ:
        raw_file = raw_file.replace(f"$ENV:{var_name}", os.environ[var_name])
```

So `$ENV:ppe_lakehouse` in `parameter.yml` is replaced with the value of the OS env var `ppe_lakehouse`. Tokens whose environment variable is not set are left unchanged (preserving prior behavior), and replacement still only happens when the feature flag is enabled.

## Behavior change

This is an intentional behavior change — anyone who literally named env vars `$ENV:...` would no longer match — but since the previous behavior was non-functional, it ships as a bug fix.

## Changes

- **`src/fabric_cicd/_parameter/_utils.py`** — Rewrote the resolver to use the plain OS variable name; kept debug logging.
- **`docs/how_to/parameterization.md`** — Reworded the *Environment Variable Replacement* section to clarify `$ENV:` is only the in-file token prefix and the OS/pipeline variable is looked up by its plain name, with a concrete `ppe_lakehouse`/`prod_lakehouse` example.
- **`tests/test_parameter_utils.py`** — Updated existing tests to use plainly-named OS env vars and added tests for: a token left untouched when its env var is missing, and multiple/repeated tokens.

## Validation

- `pytest tests/test_parameter_utils.py -k "replace_variables_in_parameter_file or replace_env_variables_in_content" -q` → **5 passed**
- `ruff format --check` and `ruff check` → clean


shirasassoon:
@copilot review the PR

src/fabric_cicd/_parameter/_parameter.py
```diff
@@ -1124,6 +1129,10 @@ def _validate_dynamic_replacement_variables(self) -> tuple[bool, str]:
         if has_cross_workspace_variables:
             logger.warning(constants.PARAMETER_MSGS["cross_workspace_variable_warning"])
 
+        # Warn if environment variables are detected but the feature flag is not enabled
+        if has_environment_variables and "enable_environment_variable_replacement" not in constants.FEATURE_FLAG:
+            logger.warning(constants.PARAMETER_MSGS["environment_variable_feature_warning"])
```
ayeshurun: dont we want to raise error?

src/fabric_cicd/_parameter/_utils.py
```diff
@@ -674,21 +678,26 @@ def replace_variables_in_parameter_file(raw_file: str) -> str:
...
+            if var_name not in os.environ:
+                logger.debug(f"Environment variable '{var_name}' is not set; keeping '{match.group(0)}'")
+                return match.group(0)
+
+            # Look up the plain variable name without the $ENV: prefix
+            var_value = os.environ[var_name]
+            logger.debug(f"Replaced {match.group(0)} with {var_value} in the parameter file")
+            return var_value
+
+        # Match the exact, case-sensitive in-file token prefix
```
ayeshurun, line 697: do we have a test for it? i could not find

src/fabric_cicd/_parameter/_utils.py
```diff
@@ -674,21 +678,26 @@ def replace_variables_in_parameter_file(raw_file: str) -> str:
...
+        pattern = rf"{re.escape(constants.ENVIRONMENT_VARIABLE_PREFIX)}(\w+)"
```
ayeshurun, line 698: since we accept here one+ words, can we cover this scenario (more than one word) in test?

src/fabric_cicd/_parameter/_parameter.py
shirasassoon: updated

src/fabric_cicd/_parameter/_utils.py
shirasassoon: Add test: `test_replace_variables_in_parameter_file_prefix_is_case_sensitive`

src/fabric_cicd/_parameter/_utils.py
shirasassoon: Added test case for multi-words to `test_replace_variables_in_parameter_file`

src/fabric_cicd/_parameter/_parameter.py
```diff
@@ -1116,25 +1118,47 @@ def _validate_dynamic_replacement_variables(self) -> tuple[bool, str]:
...
                     if value.startswith(constants.ENVIRONMENT_VARIABLE_PREFIX):
-                        has_environment_variables = True
+                        reference = (f"{param_name}[{index}].replace_value.{environment}", value)
+                        environment_variable_references.append(reference)
+                        if (
+                            self.environment == "N/A"
+                            or environment == self.environment
+                            or environment.lower() == "_all_"
+                        ):
+                            active_environment_variable_references.append(reference)
```
ayeshurun, line 1126: can you please explain this change?

src/fabric_cicd/_parameter/_parameter.py
```diff
@@ -1116,25 +1118,47 @@ def _validate_dynamic_replacement_variables(self) -> tuple[bool, str]:
...
-        if has_environment_variables and "enable_environment_variable_replacement" not in constants.FEATURE_FLAG:
-            logger.warning(constants.PARAMETER_MSGS["environment_variable_feature_warning"])
+        # Any environment variable tokens remaining after parameter loading are unresolved
+        if environment_variable_references:
+            if "enable_environment_variable_replacement" not in constants.FEATURE_FLAG:
+                environment_errors.extend(
+                    f"{location}: {constants.PARAMETER_MSGS['environment_variable_feature_disabled'].format(value)}"
+                    for location, value in environment_variable_references
+                )
+            else:
```
ayeshurun: what would happen if active_environment_variable_references is empty?

arent we missing

```py
elif active_environment_variable_references:
 environment_errors.extend(
                    f"{location}: {constants.PARAMETER_MSGS['environment_variable_unresolved'].format(value)}"
                    for location, value in active_environment_variable_referencess
```

src/fabric_cicd/_parameter/_parameter.py
shirasassoon: We only flag an unresolved env var value for the target environment.

src/fabric_cicd/_parameter/_parameter.py
shirasassoon: I will fix it.

ayeshurun (approved):

